PR #57, opened 2018-09-14
"Catched" -> "caught"

mivanov:
"Caught" is the past tense of "catch". Please ignore if this is intentional


255kb:
Now I am ashamed 😊 
Thank you for correcting this mistake!
It will be in the next release

PR #65, opened 2018-10-04
Try to parse body for all content types

alonbhaim:
Parse body for all content types otherwise it return the defaultValue


255kb:
Thank you for the pull request 👍 I will look to integrate it in v1.2.0

PR #103, opened 2019-03-10
Pretty-print JSON bodies

vankeer:
I was missing the feature to "pretty-print" a copy/pasted JSON body within Mockoon. So I added it on click of the "body" icon, because it was one of the things I tried clicking on when looking for this feature.


255kb:
Hi, 
thank you very much for the contribution! I merged it in the `next` branch. However, don't expect the next release anytime soon. I am finishing a big refactoring of the application and I have also a lot of things coming in my personal life next month, so the next release may not be available before spring / summer (I hope spring 😄)

PR #155, opened 2019-09-08
Implement logs for responses

### LINKED ISSUE #138: Allow view proxy responses
It can be very helpful can view the proxy responses from the environment log. Today I need to look this in my application logs, if I can look this on the environment logs can be more fast and easier.

Add the headers and the body of the reponse in environment logs, where the request are showed, can be solve this.

gabrielperes97:
**Description**
I create two tabs in environment logs view: Request and Response. 
Each call have your request and response data logged in these tabs. 

**Related Issue**

This pull request implement the feature described in #138

**Motivation and Context**

This is described in #138

**How Has This Been Tested?**

I used the default Dev env environment and made all calls with firefox browser. 
I call /answer to test the mocked responses, /test to test the 404 error responses.
I create another environment with a /test route with the response (200, {}, timeout 5000), using 3001 port. I enable the Dev env proxy setting to the another environment (http://127.0.0.1:3001/). When I call /test on Dev env I can see the message "No response yet", and after 5 seconds I can see the response.
I stopped the other environment and called /test again, so I can see the response when I have an proxy error.

The code changes affect the express and proxy error messages (gateway timeout, route not found, internal error), I make handlers to can log these responses, so I override the error responses.

**Screenshots**

Screenshot_20190908_161410

**Closing issues**

closes #138

### DIFF
--- a/src/app/components/environment-logs.component.html
+++ b/src/app/components/environment-logs.component.html
@@ -38,8 +38,8 @@
         <div *ngIf="environmentsLogs[activeEnvironmentUUID][(selectedLogIndex$ | async)] as selectedLog"
           class="environment-logs-content pl-2 pr-2">
           <ngb-tabset #t="ngbTabset">
-            <ngb-tab id="tab-request" title="Request">
-              <ng-template ngbTabTitle><i class="material-icons">call_made</i></ng-template>
+            <ngb-tab id="tab-request">
+              <ng-template ngbTabTitle>Request <i class="material-icons">call_made</i></ng-template>
               <ng-template ngbTabContent>
                 <div class="environment-logs-content-title" (click)="generalCollapsed = !generalCollapsed">
                   <i *ngIf="!generalCollapsed" class="material-icons">unfold_less</i>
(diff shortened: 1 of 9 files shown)


255kb:
Thank you @gabrielperes97 for this contribution. I am currently in vacation but will be back soon and will review this PR so it can make it for v1.5.0

255kb:
I really like your attention to details 👍 
Some thought after having a quick look (but I will look closer soon):
- I would invert the two arrows in the tabs. I know that a request is usually going "out" but here the request is entering Mockoon, so I would use a "down" arrow, and a "up" arrow to symbolize the response going out. What do you think?
- I would store the current tab in the store like the other tabs/views, so we have greater control on it when / if needed. main "view" and route tabs are already stored https://github.com/mockoon/mockoon/blob/3b273d2944a5c1e3ddc96c2657dc8f66579e998f/src/app/stores/store.ts#L22

gabrielperes97:
Done.

255kb:
Thank you very much for this great contribution, it will be in the next release  👍

PR #197, opened 2019-10-31
Generate a unused port to a new environment

### LINKED ISSUE #192: Automatically assign port number when creating an environment
When importing / creating / duplicating a new environment the port number is always 3000 by default. 

List all the port already assigned and chose an unused one.

gabrielperes97:
**Description**

On create or duplicate an environment it's generate a new unused port to it. The port is chosen based on the active environment port. If any of the ten next ports is unused by an environment, this port is chosen. Else a random unused port is generated. The default port, when there's no environments, remains the 3000.

**Related Issue**

Issue #192

**How Has This Been Tested?**

I tested creating a environment if when there's no environments. Creating and duplicating more than ten environments selecting the first on any new environment.

image

**Closing issues**

Closes #192

### DIFF
--- a/src/app/stores/store.ts
+++ b/src/app/stores/store.ts
@@ -173,41 +173,20 @@ export class Store {
   }
 
   /**
-   * Generate a unused port to a new environment
-   */
-  public getNewEnvironmentPort(): number {
-    const usedPorts: number[] = [];
-    this.store$.value.environments.forEach((environment) => {
-      usedPorts.push(environment.port);
-    });
-
-    const activeEnvironment: Environment = this.getActiveEnvironment();
-    let testSelectedPort: number;
-    if (activeEnvironment == null) {
-      testSelectedPort = 3000;
-    } else {
-      testSelectedPort = activeEnvironment.port;
-    }
-    for (let i = 0; i < 10; i++) {
-      if (!usedPorts.includes(testSelectedPort)) {
-        return testSelectedPort;
-      }
-      testSelectedPort++;
-    }
-
-    const min = Math.ceil(1024);
-    const max = Math.floor(65535);
-    do {
-      testSelectedPort = Math.floor(Math.random() * (max - min)) + min;
-    } while (usedPorts.includes(testSelectedPort));
-
-    return testSelectedPort;
-  }
-
-  /**
    * Update the store using the reducer
    */
   public update(action: Actions) {
     this.store$.next(environmentReducer(this.store$.value, action));
+  }
+
+  /**
+   * Get a list with all environment ports
+   */
+  public getEnvironmentsPorts(): number[] {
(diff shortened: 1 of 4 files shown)


src/app/stores/store.ts
```diff
@@ -172,6 +172,38 @@ export class Store {
       .find(response => response.uuid === this.store$.value.activeRouteResponseUUID);
   }
 
+  /**
+   * Generate a unused port to a new environment
+   */
+  public getNewEnvironmentPort(): number {
```
255kb: I would move this in the EnvironmentService class as a private method. Store should stay very basic with no business logic I think.

src/app/stores/store.ts
```diff
@@ -172,6 +172,38 @@ export class Store {
...
+    });
+
+    const activeEnvironment: Environment = this.getActiveEnvironment();
+    let testSelectedPort: number;
+    if (activeEnvironment == null) {
+      testSelectedPort = 3000;
+    } else {
+      testSelectedPort = activeEnvironment.port;
+    }
+    for (let i = 0; i < 10; i++) {
```
255kb: This make sense if we want to continue the serie 3000, 3001, 3002. But if one chose the port 65535, it will increase to 65536 and so on. I know there is no limit on the input field but we should add a check to avoid this. Or directly use the random method below.

test/environments-create-delete.spec.ts
```diff
@@ -29,4 +30,14 @@ describe('Create and delete environments', () => {
 
     await tests.spectron.client.waitForExist('.header input[placeholder="No environment"]');
   });
+
+  it('Add ten environments ever clicking in the first', async () => {
+    let port = 3000;
+    for (let i = 0; i < 10; i++) {
```
255kb: You could reduce the code a little bit and directly use `port` as counter variable
```suggestion
    for (let port = 3000; port < 3010; port++) {
```

test/lib/helpers.ts
```diff
@@ -132,4 +132,9 @@ export class Helpers {
   async httpCallAsserter(httpCall: HttpCall) {
     await this.httpCallAsserterWithPort(httpCall, 3000);
   }
+
+  async assertActiveEnvironmentPort(expectedPort: number) {
+    const val: String = await this.testsInstance.spectron.client.element('input[formcontrolname="port"]').getAttribute('value');
```
255kb: Please use a more meaningful variable name like `port` instead of `val`.

gabrielperes97:
Done.

255kb:
Thank you @gabrielperes97 this is a great addition :) 
Everything will be in v1.6.0. However, I am debugging a serious issue where the application is not loading anymore when packaged. Apparently this is linked to Electron 6. So I am not sure when the v1.6.0 will be out.

PR #198, opened 2019-10-31
Feature/variable log body size limit

### LINKED ISSUE #186: Variable log's body limit size
**Is your feature request related to a problem?*
I work a much with large jsons and many times i'm frustrated with the logs request/response body truncated.

On the Application Settings view, we can put a input field to set the log's body size limit to truncate. 

The input field can be placed in the environment settings view. 
Or we can use a button to disable the log's body truncate.

gabrielperes97:
**Description**

Implements a variable log body size. On the Settings modal we can setting the value for truncate the log's request and response body. The default value is 10000.

**Related Issue**

Issue #186 

**How Has This Been Tested?**

I enabled the proxy to a mirror api and send large random text. Next I change the variable value and made the request again.

image

**Closing issues**

Closes #186

### DIFF
--- a/src/app/components/settings-modal.component.html
+++ b/src/app/components/settings-modal.component.html
@@ -8,6 +8,14 @@
       <label class="custom-control-label" for="analytics">I agree to share anonymous usage data</label>
     </div>
     <p class="text-muted"><small>We are using a basic Google Analytics setup, none of your data/content is kept anywhere and it helps us improve Mockoon :)</small></p>
+
+    <div>
+      <label for="log-body-size">Log body size:</label>
+      <input type="number" class="form-control col-4" id="log-body-size"
+          ngbTooltip="Log's body size" min="1"
+          placeholder="10000 (default)" digitOnly 
+          [ngModel]="(settings$ | async)?.logSizeLimit" (ngModelChange)="settingsUpdated($event, 'logSizeLimit')">
+    </div>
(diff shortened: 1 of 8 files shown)


src/app/components/settings-modal.component.html
```diff
@@ -8,6 +8,14 @@ <h4 class="modal-title">Settings</h4>
       <label class="custom-control-label" for="analytics">I agree to share anonymous usage data</label>
     </div>
     <p class="text-muted"><small>We are using a basic Google Analytics setup, none of your data/content is kept anywhere and it helps us improve Mockoon :)</small></p>
+
+    <div>
+      <label for="log-body-size">Log body size:</label>
```
255kb: Could you add a small explanation (with the unit, bytes?) after the label? Like this:
`<p class="text-muted"><small>Truncate the entering requests body if bigger (bytes).</small></p>`

gabrielperes97: I add "Truncate the entering requests and exiting responses bodies if bigger (characters)." because it's limit the size of response body log too. What you think?

255kb:
Thank you @gabrielperes97

src/app/components/settings-modal.component.html
255kb: This is perfect, even clearer :)

PR #204, opened 2019-11-09
173 collapse environments menu

### LINKED ISSUE #173: Add possibility to collapse environment menu
Environment menu may not be used as frequently as the rest of the interface. Majority of users only have 1 environment.

Allow users to collapse the environment menu

JoshMLeslie:
**Description**

1 - refactored environment menu code into its own component / module
2 - added ability to collapse environment menu, which saves state to settings file

**Related Issue**

closes #173 

**Motivation and Context**

closes #173

**How Has This Been Tested?**

By hand. Some testing was added to the test suite but was unable to run locally. Will update.

### opened
image

### closed
image

**Closing issues**

closes #173


JoshMLeslie:
@255kb I had a feeling this would throw a bunch of tests out of whack. I've been meaning to ask, are you developing on a Linux OS ?

255kb:
Sorry for the delay @JoshMLeslie and good job. It looks great. 
Now as I am looking at the screenshots I wonder if it wouldn't be better if we still show the environments even when collapsed. Most menu like that have an expanded state with icon and text, and only icons when collapsed. 
Maybe it could be still useful to see an entry for each environment (the first letter of the environment name or something like this) with the three icons (cors, https, proxy). And like that we would still be able to easily switch from one environment to the other and right click on it. 
What do you think?
Concerning the tests I expected them to fail. If you want, I can take your branch and correct the tests when feature is finished.

JoshMLeslie:
### new side menu
image

### expanded menu
image

JoshMLeslie:
@255kb All set for you 👍

255kb:
thank you @JoshMLeslie it looks nice! 
I will take it from here to correct the tests. But I have very little time at the moment and my priority is to finish the openAPI import export feature. So I am not sure when I will be able to work on this.

255kb:
@JoshMLeslie I resolved the conflicts and merged your PR on a feature branch. Like this it will be easier for me to update the tests.

JoshMLeslie:
@255kb Sounds good! Thanks for taking it the rest of the way.

255kb:
Hi @JoshMLeslie, I wish you a happy new year 🎉 
I want to apologize because instead of only adding the tests to your code I also refactored what you did. 😳 
I liked it, but it felt like a little bit weird to have two menus, one hovering the other.  So I reverted back to animating the width of the menu while keeping the look and feel of what you did. I think the final result is quite good and performant enough. I must admit that it was a real pain to have something nice, performant, good enough UX wise, etc., and I am grateful that you did most of this hard work!
Also, I removed the menu state from the settings file. My bad. I asked you to do this, but after thinking about it, I think the user has no interest in saving the state of a menu which hides part of the interface.
Finally, I removed the module you created for the environment menu component. Not because I am against modules (😄) but I think that having only the root module is enough for now. It's a monolithic app, with less than ten component, that will be only used once (for most of them), and cannot be lazy loaded. 
Maybe one day we will reorganize in different modules, but I prefer keeping things very simple for now and avoid any unnecessary architecture. But I am sure we could debate this a lot and I may be mistaken. 

I also took this opportunity to extract the routes menu in a separated component, reorganize the styles a little bit, extract some scrolling methods in a service, etc. 
Things I wanted to do since a long time :) 

Thank you again for your work, I really appreciate it!
I don't know when this will be released, probably not before 1 or 2 months (max) as I also want to finish the openApi import / export (I am "working" on it since months now...)

PR #227, opened 2019-12-11
feat(226): Add setting Truncate route name in list

johnwest80:
**Description**

Provide a setting to determine whether or not the route name should be truncated in the route list

**Related Issue**

None

**Motivation and Context**

When you have many routes that start with the same text, it's hard to know which route is the one you're looking for

**How Has This Been Tested?**

Locally

image

image

**Closing issues**

#227


255kb:
Thank you for your contribution! I corrected conflicts and merged :)

Issue #136, opened 2019-07-20
404 page not adding custom headers


skunkworker:
**Describe the bug**
When custom headers are present for a given environment, navigating to a undefined route will show a 404, but this won't include the custom headers that were set beforehand.

I ran into this issue when using https://github.com/axios/axios. I was getting a invalid CORS error for the mocked server, and hunting it down I was able to see that a 404 wasn't including the custom headers.

**To Reproduce**
Add custom headers to an environment.
Navigate to a route that isn't recognized, throwing a 404.

**Expected behavior**
All routes, including invalid ones should return the same custom headers.

**Mockoon version:**
1.4.0

**OS / OS version:**
Mac OS 10.13.6

abiodunjames:
@255kb I would like to pick this. Is that fine for you?

255kb:
Sure @abiodunjames 
Solution shouldn't be very complicated I think.
Just be careful when handling 404 with ExpressJS, that the proxy mode (and other things) still works. There are tests but not everything is covered yet. So we still need to be careful and test a lot 🙂
And if you can add tests also it would be great (we only add automated tests for now, you can see examples in test folder)
Thank you!

PR #223, opened 2019-12-10
Custom proxy request/response headers

neocodegs:
**Description**
Support adding custom request and response headers for proxy mode

**Related Issue**

**Motivation and Context**

**How Has This Been Tested?**
 Test for headers properties presence
proxyHeaders

**Screenshots (if appropriate):**
Snapshot_2019-12-03_16-19-46

**Closing issues**

PR #232, opened 2020-01-02
Fix/issue-136: Always return environment headers in responses

abiodunjames:
**Description**
This PR ensures that  headers defined in proxy mode are always returned in responses.

**Related Issue**

https://github.com/mockoon/mockoon/issues/136

**Motivation and Context**
In my projects, I often use custom headers like `x-auth-key` for authentication. If you use Mockoon with this header (x-auth-key) set in an environment, it will work if the proxied server returns a  200 response.  Incase of a 404 error from the proxied server, the header `x-auth-key` is ignored. See https://github.com/mockoon/mockoon/issues/136

**How Has This Been Tested?**

* Use Mockoon in proxy mode with some headers defined. 
* Access some valid endpoints in the proxied server from your browser and inspect  the response headers. It should return the headers defined
* Access an endpoint that does not exist in the proxied server such that you get a 404 error.  The headers you defined in mockoon should be returned in the response as well.

** Unit/UI test?**
No, this is my entry PR to this project. Things are still sketchy for me.

### DIFF
--- a/src/app/services/server.service.ts
+++ b/src/app/services/server.service.ts
@@ -71,6 +71,7 @@ export class ServerService {
 
     // apply latency, cors, routes and proxy to express server
     this.logRequests(server, environment);
+    this.setCustomHeaders(server, environment);
     this.logResponses(server, environment);
     this.setEnvironmentLatency(server, environment.uuid);
     this.setRoutes(server, environment);
@@ -346,6 +347,23 @@ export class ServerService {
       const enhancedReq = req as IEnhancedRequest;
       enhancedReq.uuid = log.uuid;
       this.store.update(logRequestAction(environment.uuid, log));
+      next();
+    });
+  }
+
+  /**
+   * Ensure that environment headers are always set in responses
+   *
+   * @param server - the server serving responses 
+   * @param environment - the environment where the headers are configured
+   */
+  private setCustomHeaders(server: any, environment: Environment) {
+    server.use((req, res, next) => {
+      for (const header of environment.headers) {
+        if (!res.hasHeader(header.key)) {
+          res.setHeader(header.key, header.value);
+        }
+      }
       next();
     });
   }


255kb:
@abiodunjames If you define a middleware that add environment headers, I think you should remove it from route declaration (https://github.com/mockoon/mockoon/blob/ac1c2faa77727853e70aa12a0622a15d9d5c4f98/src/app/services/server.service.ts#L171). 
Also, be careful about the order when applying the headers. A route response header should take precedence over an environment "global" header. 
Anyway, a new test file should be added to test multiple cases (and things that are not yet tested): 
- route response headers
- environment headers
- precedence of route headers over environment headers
- environment headers present on proxied calls
- environment headers present on proxied calls that didn't match any of the proxied server routes (return 404)

I can help with tests at one point, if you need. 
Thank you!

255kb:
I completely forgot, but you may want to build your solution on this pull request https://github.com/mockoon/mockoon/pull/223/files which introduce proxy headers (headers that are only transmitted to the proxied service, or added to responses from the proxy service). 
Maybe some code could be shared / reused.
I didn't review this pull request yet. I will do it probably in the next 2 weeks.

abiodunjames:
> I completely forgot, but you may want to build your solution on this pull request https://github.com/mockoon/mockoon/pull/223/files which introduce proxy headers (headers that are only transmitted to the proxied service, or added to responses from the proxy service).
> Maybe some code could be shared / reused.
> I didn't review this pull request yet. I will do it probably in the next 2 weeks.

I  agree. It makes sense to build my solution on that. I'll wait for your review

255kb:
@abiodunjames #223 has been merged, it brings many changes. You may want to rebase and continue your work from latest master

abiodunjames:
> @abiodunjames #223 has been merged, it brings many changes. You may want to rebase and continue your work from latest master

Thanks @255kb.  I created a new PR https://github.com/mockoon/mockoon/pull/240
Therefore I'm closing this in favour of https://github.com/mockoon/mockoon/pull/240

PR #236, opened 2020-01-06
Add cookie to dummy helpers (closes #235)

### LINKED ISSUE #235: Add cookie to dummy helpers
I wanted to use a value from a cookie in my response but the cookie header is formatted and may contain multiple cookies so I can't trivially pull the value of a given cookie key.

I'm proposing to add `cookie key default` to the dummy helpers

sleepypikachu:
**Description**
Add cookie dependency to project.

Use cookie library to parse the value of the cookie header (if present)
and then extract the value of the requested key (if present).

Returns defaultValue in the pattern of the other dummy helpers if the
cookie is not set/no cookie header was provided.

**Motivation and Context**
> I wanted to use a value from a cookie in my response but the cookie header is formatted and may contain multiple cookies so I can't trivially pull the value of a given cookie key.

**How Has This Been Tested?**

Manually executed requests with header present with cookie, no header, header without the cookie value to get the value, default, default as expected.

**Closing issues**
Closes #235


sleepypikachu:
@255kb don't think this test failure is anything to do with my change. Is the test flaky? https://github.com/mockoon/mockoon/pull/236/checks?check_run_id=377148047#step:5:214

255kb:
@sleepypikachu yes this test is flaky since some days! I don't know why, I will investigate.

255kb:
Thank you for the contribution! I think I will just replace `cookie` library by `cookie-parser` which is an expressJS middleware and could be used in other parts of the app later (rules based on cookies, etc)

PR #242, opened 2020-01-20
Insert duplicates after originals (closes #239)

### LINKED ISSUE #239: Duplicated route located after the original
When you duplicate a route, it should be good to have the cloned one right after the original, not at the end of the list.

I use to organize the routes with the same path together, but when I duplicate a route I have to go down to the bottom of the list of routes and move it. It is tedious.

sleepypikachu:
**Description**

Adds an optional "afterUUID" string to each of the addX actions and
splice the clone into the right place in the list in the reducer if
afterUUID is specified.

Ignore vim swp files.

**Motivation and Context**

When duplicating routes and environments, insert the duplicate after the
original in the list (rather than at the end of the list). Saves the user from having to drag the duplicate from the bottom of the list.

**How Has This Been Tested?**

Add test coverage on duplicating routes and environments that covers the
insertion index. This could be extended to cover faithful duplication
but that's outside the scope of this MR.

Manual testing in development mode.

**Closing issues**
Closes #239


255kb:
Thank you for your contribution @sleepypikachu 🚀

Issue #104, opened 2019-03-22
Allow resizing route column


DevRyanDonahue:
**Is your feature request related to a problem? Please describe.**
When the routes in my environment become long, it can become impossible to differentiate which is which based on the routes list. The last of the text is truncated so I have to navigate to each route to see what it is

**Describe the solution you'd like**
Allowing resizing of this column would likely be the best solution. A hover tooltip on route items would be a next best.

**Describe alternatives you've considered**
Truncating at beginning of text or the middle, but they could still hold valuable information that making the column bigger could still display.

**Additional context**

axc450:
I can look into adding this, but it might not make it into the next release

sotcha:
Hi, I have a close to final implementation for this feature. Do you want to get a PR for this ?

255kb:
Hi @sotcha . All contributions are welcome! 
However, I would like to discuss the implementation with you first. How do you plan to do this? Like a traditional resize (mouse click on the border, hold, move, release click)? As it may leads to lots of repaint, because of column resize, how are the performances? 
Also, On the master branch (upcoming version), I worked on a collapsible environments menu with lots of refactoring. You absolutely **must** start your work from this branch :)

sotcha:
Hi @255kb , yes I forked from the master branch.

Well, I created a directive which works (as you described like the traditional resize) :

- Start when onMouseClick on the directive
- Register _mousemove_ and _mouseup_ listener on the body
- Release on mouse up

The performance seems to be OK, at least on my laptop.

I have a draft commit here : 
https://github.com/sotcha/mockoon/tree/feature/allow-resizing-route-column-%23104

Have a look if you want.

255kb:
This looks great! Feel free to open the PR. I will try to review it in the coming days/weeks. Anyway, it can make it for the next release. I have still things I am working on and bugs I want to solve for 1.8.0.

sotcha:
Hi @255kb , what do you think about adding the same functionality into the environment logs column.

Something like this 
![image](https://user-images.githubusercontent.com/1183265/76159496-bff0bc80-6129-11ea-9920-02ca6554ab82.png)
 

Also as I can see into the description it is we would like to add a hover tooltip over the url name 

![image](https://user-images.githubusercontent.com/1183265/76159580-e4996400-612a-11ea-9ba0-2927acde0159.png)

Should I add these to this PR?

255kb:
[...]

PR #264, opened 2020-03-09
Added - #104: Allow resizing route column

sotcha:
**Description**

Added _ResizeColumnDirective_ which : 
- Tracks **mousedown** event
- Adds **mousemove** and **mouseup** to document.body in order to calc the new width and apply it (with css of parent element).

The directive added to 
* _src/app/components/environment-logs.component.html_
* _src/app/components/routes-menu/routes-menu.component.html_

Also, tooltips added to these files for the route and environment logs urls.

**Related Issue**

Issue : #107 

**Motivation and Context**

**How Has This Been Tested?**

It is tested manually :
- Start the server 
- Place the mouse to the routes column right edge and drag it.

_ps : I am not sure how we can write tests for this_

image

**Closing issues**
closes #107


src/app/directives/resize-column.directive.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.initResizableColumns();
+  }
+
+  /**
+   * Registers event handlers to body
+   * - mousemove: mark as pressed, calc and apply the new width
+   * - mouseup: mark as non-pressed
+   */
+  private initResizableColumns() {
+    this.renderer.listen('body', 'mousemove', (event) => {
```
255kb: Could you instead use `@HostListener`?

src/app/directives/resize-column.directive.ts
```diff
@@ -0,0 +1,94 @@
...
+        } else if (width > maxWidth) {
+          width = maxWidth;
+        }
+
+        // Apply the new width
+        ResizeColumnDirective.applyWidthCss(this.startElement.parentElement, width);
+      }
+    });
+
+    this.renderer.listen('body', 'mouseup', (event) => {
```
255kb: Could you instead use `@HostListener`?

src/app/directives/resize-column.directive.ts
sotcha: Well, I can use `@HostListener`, but it will listen continuously. I think we should register-unregister, which is not supported by HostListener, correct ?

255kb: True, didn't think about this. You can keep it like this 👍 
And yes Angular doesn't have conditional HostListeners https://github.com/angular/angular/issues/7626

sotcha:
> It looks great, and performances seem to be good.
> Two things I would like to do differently:
> 
> * it would be safer to have hard pixel limits for min and max size that you can pass as the directive config/params. Something like 200-500px for the route menu (which is already very big), and 200-400px for the environment logs. Otherwise, layout can be completely broken due to flx/Bootstrap. Anyway, this will need to be tested thoroughly after setting up these hard limits.
> * these settings would need to be save in the settings file. This is not very complicated. Idea is to:
>   
>   1. add the properties to the `Settings` type (routeMenuSize, envLogsSize),
>   2. add some code to "migrate" the settings in the `migrateSettings` method in settings service,
>   3. call the `updateSettings` method as soon as mouse is released,
>   4. after that settings will be saved automatically
> 
> Thank you!


* Well for the hardcoded min-max size, I added params to the directive
` <resize-column [type]="'routeMenu'" [minWidth]="200" [maxWidth]="500"></resize-column>` 
I think that setting a min is ok, but I am not sure if setting max hardcoded is right.

Let's say that a user has a wide screen and resizes mockoon to be wide, I do see the reason to limit ie _routes menu_ to **500px** .

image

This is an example where _routes menu_ can be bigger.

On the other hand, we can see that the UI looks broken when I resized the window. 

image

I think that since the user can manually resize the window, this is acceptable.


* Also for the settings, I added 
```
  routeMenuSize: number;
  envLogsSize: number;
```
for saving the current size, do but we want for the min-max ?

255kb:
Saving current sizes for both menus is definitely enough. 
Concerning the max size, I understand, but I would like to avoid people being able to "break" the interface. It will surely back fire into GitHub issues or feeling that UI is not always good. 
Maybe we could set a hardcoded minimum like 200px and dynamic max like you did before? But instead of allowing 80% it could be 40% or something like that. 
But allowing resizing to nearly the whole width, which completely breaks the interface is unacceptable I think, from a user perspective. Also, most softwares I know have a min/max size for resizable items.

sotcha:
Hi @255kb , 
Well, I found a better idea (after some tests) to set a minimum width of the right area since this is the element that we need to keep it not compressed. 

`<resize-column [type]="'routeMenu'" [minWidth]="150" [minWidthFromRight]="640"></resize-column>`

And I kept the rule with the max width based on body and a factor, in case that there is no _minWidthFromRight_ passed.


Again, I do not think that solves the issue 100% (the UI in some cases breaks somehow), but seems a close enough solution. I would like to hear your opinion.

255kb:
Thank you for the modification. 
I am sorry for the late reply, I was quite busy debugging the last version. 
I am reviewing your work right now and may probably ask for some modification. Let me know if you don't have more time to spend on this. I can also take it from here if needed 👍

src/app/services/settings.service.ts
```diff
@@ -29,7 +31,9 @@ export class SettingsService {
     analytics: true,
     bannerDismissed: [],
     logSizeLimit: 10000,
-    truncateRouteName: true
+    truncateRouteName: true,
+    routeMenuSize: undefined,
+    envLogsSize: undefined
```
255kb, line 36: Please rename to `envLogsMenuSize` or `logsMenuSize` for consistency.

src/app/directives/resize-column.directive.ts
```diff
@@ -0,0 +1,144 @@
+import {ElementRef, Renderer2, HostListener, Directive, AfterViewInit, Input} from '@angular/core';
+import {SettingsService, SettingsProperties} from '../services/settings.service';
```
255kb, line 2: please use absolute import `src/...`

src/app/directives/resize-column.directive.ts
```diff
@@ -0,0 +1,144 @@
...
+  // tslint:disable-next-line:directive-selector
+  selector: 'resize-column'
+})
+export class ResizeColumnDirective implements AfterViewInit {
+
+  constructor(private elementRef: ElementRef,
+              private settingsService: SettingsService,
+              private store: Store,
+              private renderer: Renderer2) {
+    elementRef.nativeElement.style.cursor = 'col-resize';
```
255kb, line 24: This should be moved to the styles.scss file https://github.com/mockoon/mockoon/pull/264/files#diff-354873b4b545926f38dc09efd5af2b6dR315

src/app/directives/resize-column.directive.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+
+  /**
+   * Registers event handlers to body
+   * - mousemove: mark as pressed, calc and apply the new width
+   * - mouseup: mark as non-pressed
+   */
+  private initResizableColumns() {
+    // this.renderer.listen('body', 'mousemove', (event) => {this.handleMouseMoveEvent(event)});
```
255kb, line 75: remove comment

src/app/directives/resize-column.directive.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (this.type === 'routeMenu') {
+        width = settings.routeMenuSize;
+      } else {
+        width = settings.envLogsSize;
+      }
+
+      if (typeof width !== 'undefined') { // finally update width if needed
+        this.applyWidthCss(width);
+      }
+    }
```
255kb, line 57: We could eventually have a race condition where settings are not yet loaded. 
Please use something like this instead:

```suggestion
    this.store.select('settings').pipe(
      filter(settings => !!settings),
      first(),
    ).subscribe((settings) => {
      let width;
      if (this.type === 'routeMenu') {
        width = settings.routeMenuSize;
      } else {
        width = settings.envLogsSize;
      }

      if (typeof width !== 'undefined') { // finally update width if needed
        this.applyWidthCss(width);
      }
    });
```

sotcha:
@255kb no worries. I can spend time on this (maybe not immediately, but I have).

src/app/directives/resize-column.directive.ts
sotcha: Sorry for that

255kb: no need to apologize :)

sotcha:
So if we 
- listen window resize 
- maxWidth = document.body.offsetWidth * this.maxPercentageSize / 100

I think there is no point keeping the _minWidthFromRight_ limitation, correct?

255kb:
Yes I think so too. Having a maximum viewport % should be enough. 
I did not forget your PR, I will review it in the coming day or two, and I apologize! 🐱

255kb:
Thank you @sotcha it's working like a charm! 
One thing that I would like to modify is to invert the route menu and logs menu initial sizes and maxwidth factors. Routes menu is definitely more important and should be able to take more space than environment logs. 

If you agree I can take it from here, do this small change, resolve the conflict and merge.

sotcha:
@255kb Yes please continue it. I am glad that I helped with that. :fireworks:

255kb:
@sotcha It's merged! I just fixed one small thing and inverted the sizes. 
Thank you again for this great contribution! 🥳

PR #273, opened 2020-03-20
201: add options "or" & "and" to first rule

### LINKED ISSUE #201: Add "AND" relations in route responses rules
Currently I saw user can add multiple rules for a response. But these rule check condition is "OR", for example if I have 2 rules for this response, then if rule1=true or rule2=ture, then this response will match and returned. Whether I want the AND relation, which means only when rule1=true AND rule2=true, this response will return. Do we have this feature? if not, when?

ninjaneer164:
**Description**
added options "or" & "and" (radio buttons) to first rule.

**Related Issue**
https://github.com/mockoon/mockoon/issues/201

**Motivation and Context**
in some cases, we need a particular response to match when ALL rules pass, not "at least one".

https://github.com/mockoon/mockoon/issues/201

**How Has This Been Tested?**
this has been tested in an actual project work by multiple developers.
the change only affects "response matching" logic.

Screen Shot 2020-03-20 at 5 22 07 AM

**Closing issues**
closes #201


src/app/classes/response-rules-interpreter.ts
```diff
@@ -67,7 +75,7 @@ export class ResponseRulesInterpreter {
       return value.includes(rule.value);
     }
 
-    return value === rule.value;
+    return value.toString() === rule.value.toString();
```
ninjaneer164: apparently, when an env is exported, numbers get converted to string, i.e. ("foo": "1"). when another dev imported the json, the rule no longer worked. i'm using toString() - not a fan of "==" - to ensure that numbers work as rule as well.

src/app/classes/response-rules-interpreter.ts
```diff
@@ -27,11 +27,19 @@ export class ResponseRulesInterpreter {
    * If no rule has been fulfilled get the first route response.
    */
   public chooseResponse(): RouteResponse {
-    return (
-      this.routeResponses.find(routeResponse => {
-        return !!routeResponse.rules.find(this.isValidRule);
-      }) || this.routeResponses[0]
-    );
+    let response = this.routeResponses.find(routeResponse => {
+      const andRules = routeResponse.rules[0].andRules;
```
ninjaneer164: or/and options only appear with first rule, so only need to look at [0]

src/app/classes/response-rules-interpreter.ts
```diff
@@ -27,11 +27,19 @@ export class ResponseRulesInterpreter {
...
+
+      return andRules
+        ? !!routeResponse.rules.every(this.isValidRule)
+        : !!routeResponse.rules.find(this.isValidRule);
+    });
+
+    if (response === undefined) {
+      response = this.routeResponses[0];
+    }
```
ninjaneer164: if no match found, use first one as default

src/app/components/route-response-rules.component.ts
```diff
@@ -65,7 +69,12 @@ export class RouteResponseRulesComponent implements OnInit {
     }
 
     newRules.forEach(rule => {
-      formRulesArray.push(this.formBuilder.group(<ResponseRule>{ ...rule }));
+      formRulesArray.push(
+        this.formBuilder.group(<ResponseRule>{
+          ...rule,
+          andRules: rule.andRules || false
```
ninjaneer164: default andRules to false, rather than leaving it undefined

krishnaSpotnana:
@255kb merging any time soon?

255kb:
@krishnaSpotnana I cannot give an ETA. I have a lot ongoing already for the next release, and I need to review this and work with @ninjaneer164 to finalize it. It will definitely be in the next release but it usually takes a couple of months as I can only spent a certain amount of time on this project (it's a side project).

255kb:
@ninjaneer164 Sorry for getting back to you so late, I wanted to finish other task before working with you on this. 
First I have a question: Does this solve your needs? I would have expected an **and/or** switch *between* each rules so you can have more complex use cases like `rule AND (rule OR rule)`.
I understand that your implementation is rather a global switch at the route response's level.
I would say that this may be a good first step. However in this case, it would require some changes: 
- the new `andRules` property should be on the `RouteResponse` object and not the `ResponseRule`.
- radio button should be move to a different place so we understand that it applies to all the rules and not just the first one. 

In any case, automated tests should be written for this switch (`./test` folder), a migration written to add the new property to all the route responses, and updates done accordingly to all the tests `environments.json` files in the test folder. But I can do all these tests related tasks as it is rather tedious 😄

255kb:
@ninjaneer164 I will take it from here, keep the logic you added but modify the UI a little bit and add the missing tests and migrations. Thank you for your contribution!

Issue #250, opened 2020-02-04
Allow duplication of route responses


255kb:
And one more improvement which I find useful is to create a possibility to duplicate response 
![image](https://user-images.githubusercontent.com/21062077/70854008-447ac880-1eb6-11ea-89f8-f9c3295ea070.png)
, same way as existing "Duplicate route" works. 
![image](https://user-images.githubusercontent.com/21062077/70853992-10070c80-1eb6-11ea-9d8f-3a1cfed55437.png)
At least in my projects I see it useful as I have a lot of similar responses, with small differences only

_Originally posted by @aantoniv in https://github.com/mockoon/mockoon/issues/191#issuecomment-565749413_

izerozlu:
I would like to look into that. I see that no is assigned yet. Is it available?

255kb:
Hi @izerozlu Sure, you can work on this. 
Here are some guidances, as I already thought about this feature a little bit. 
- I think the same icon as route duplication should be used. 
- it could be place next to the trash used to remove a route response
- duplication behavior should be the same than environments and routes: new duplicated route response should be inserted next to its source (and not at the end of the list). 
- "(copy)" could be appended at the end of the route response's label

This will require creation of a new Action for the reducer. 
You can take inspiration from the environment duplication methods: https://github.com/mockoon/mockoon/blob/master/src/app/services/environments.service.ts#L137

Please confirm me if you want to work on this so I can assign this issue to you. 

Thank you!

izerozlu:
@255kb You've given everything required for the further development. Thanks for that :).
Please see this reply as the confirmation.

255kb:
No problem! Do not hesitate if you have any question.

255kb:
@izerozlu I just pushed a big refactoring on master, please star your work from there, it will easier the review/merge. Thank you!

izerozlu:
Thanks for the heads up. I've not began working on the issue yet so no worries.

izerozlu:
Hey @255kb, I've began working on this issue. Mostly finished actually. Btw sorry for starting so late on the issue.

I'm curious about where to add testing for the functionality. Can you point me to some testing scenarios I could inspire from?

255kb:
No problem, there is no deadline :) 
For now there are only UI functional tests done with Spectron (Electron testing framework based on Webdriver). 
[...]

PR #306, opened 2020-06-06
Route response duplication #250

izerozlu:
**Description**

* Route response duplication button added next to route response removal button.
* `EnvironmentsService#duplicateRouteResponse` implemented with inspiration from `EnvironmentsService#addRouteResponse`.
* Route response duplication action created.
* Duplicating the active route response implemented in `AppComponent`.
* Reducer for the `ActionTypes.ADD_ROUTE_RESPONSE` enhanced to include the ability to duplicate a given route response.
* Test helper `removeRouteResponse` updated to cope with the changes.
* Tests added for route response duplication in `route-response-duplicate.spec.ts`

**Related Issue**
#250 

**Motivation and Context**

Adds the ability to duplicate a route response with a single click of a button.
#250 is the issue this feature suggested in.

**How Has This Been Tested?**

Feature tested with spectron tests. 

**Screenshots:**
Screenshot from 2020-06-06 13-26-37

Screenshot from 2020-06-06 13-26-51


src/app/app.component.ts
```diff
@@ -669,4 +684,14 @@ export class AppComponent implements OnInit {
     this.environmentsService.setActiveView('ENV_LOGS');
     this.environmentsService.setActiveEnvironmentLogTab('RESPONSE');
   }
+
+  /**
+   * Duplicate the active route response
+   */
+  public duplicateRouteResponse() {
+    this.store.selectActiveRouteResponse().pipe(first())
```
255kb, line 692: You can remove this subscription and directly use `this.store.getActiveEnvironment();` which is not an observable. Also, you can put this code in the environment service instead of the app component like there: https://github.com/mockoon/mockoon/blob/master/src/app/services/environments.service.ts#L138

src/app/app.component.html
```diff
@@ -358,6 +358,15 @@
                             </li>
                           </ul>
 
+                          <!-- Response duplication button -->
+                          <button
+                            *ngIf="!(deleteCurrentRouteResponseRequested$ | async)"
```
255kb, line 363: I think you could avoid hiding the button when a deletion has been requested but not confirmed. I don't think it's really necessary. And the user could have misclicked on the trash icon and would have to wait for the timeout in order to duplicate.

src/app/stores/reducer.ts
```diff
@@ -728,21 +729,32 @@ export function environmentReducer(
...
     case ActionTypes.ADD_ROUTE_RESPONSE: {
       const newRouteResponse: RouteResponse = action.routeReponse;
-
+      const newUUID = uuid();
+      const activeRouteResponseUUID = state.activeRouteResponseUUID === newRouteResponse.uuid ?
+        newUUID : newRouteResponse.uuid;
       newState = {
         ...state,
-        activeRouteResponseUUID: newRouteResponse.uuid,
+        activeRouteResponseUUID,
```
255kb, line 737: This part should not be changed. As for other actions, the new object (here a route response) should come with a new UUID from the service (like for env duplication https://github.com/mockoon/mockoon/blob/master/src/app/services/environments.service.ts#L154). Please create a new `renewRouteResponseUUID` method in the data service for this or modify the UUID in the schema builder method you created.

src/app/stores/reducer.ts
```diff
@@ -728,21 +729,32 @@ export function environmentReducer(
...
+                      (routeResponse: RouteResponse) => {
+                        return routeResponse.uuid === state.activeRouteResponseUUID;
+                      }
+                    );
+                    newRouteResponse.uuid = newUUID;
+                    responses.splice(activeRouteResponseIndex + 1, 0, newRouteResponse);
+                  } else {
+                    responses.push(newRouteResponse);
+                  }
+                  return { ...route, responses };
```
255kb, line 757: Please implements the same logic as the route or env duplication by providing an `afterUUID` param directly to the action (https://github.com/mockoon/mockoon/blob/master/src/app/stores/actions.ts#L235). Then you can follow the ADD_ENV implementation https://github.com/mockoon/mockoon/blob/master/src/app/stores/reducer.ts#L352

src/app/app.component.html
```diff
@@ -358,6 +358,15 @@
...
+                            type="button" id="duplication-button" class="btn btn-link btn-icon"
```
255kb, line 364: Could you please use a more specific id like `route-response-duplication`, just in case. 
Also, if you have time could you please add an ID on the deletion button too, to use it in the tests? I realize that the selector is quite cumbersome (`.this:not(.that)...`)

test/lib/helpers.ts
```diff
@@ -41,12 +42,22 @@ export class Helpers {
...
     await this.testsInstance.app.client.element(deleteButtonSelector).click();
     await this.testsInstance.app.client.element(deleteButtonSelector).click();
   }
 
+  async duplicateRouteResponse() {
+    const duplicationButtonSelector = '#route-responses-menu #duplication-button';
+    await this.testsInstance.app.client.element(duplicationButtonSelector).click();
+  }
+
+  async removeDuplicatedRouteResponse() {
```
255kb, line 56: I think this helper is a little bit too specific, please select the route response during the test using `selectRouteResponse()` and delete it using the existing `removeRouteResponse()` helper.

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,107 @@
+import { Tests } from 'test/lib/tests';
+import { expect } from 'chai';
+
+const tests = new Tests('basic-data');
+
+describe('Duplicate a route response', async () => {
+  tests.runHooks();
+
+  it('Verify single route response', async () => {
```
255kb, line 9: this test could be removed as the same test is done in the next `it()`

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    const secondRowFirstControlSelector = '#route-response-headers .row.headers-list:nth-child(2) .form-control:first-child';
+    const secondRowSecondControlSelector = '#route-response-headers .row.headers-list:nth-child(2) .form-control:nth-child(2)';
+    expect(await tests.app.client.getValue(firstRowFirstControlSelector))
+      .to.equal('Content-Type');
+    expect(await tests.app.client.getValue(firstRowSecondControlSelector))
+      .to.equal('application/json');
+    expect(await tests.app.client.getValue(secondRowFirstControlSelector))
+      .to.equal('test');
+    expect(await tests.app.client.getValue(secondRowSecondControlSelector))
+      .to.equal('test');
```
255kb, line 83: Please consider creating a small helper for this

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+      .should.eventually.have.property('value')
+      .to.be.an('Array')
+      .that.have.lengthOf(1);
+
+    const ruleTargetSelectSelector = 'app-route-response-rules .row select[formcontrolname="target"]';
+    const ruleModifierInputSelector = 'app-route-response-rules .row input[formcontrolname="modifier"]';
+    const ruleValueInputSelector = 'app-route-response-rules .row input[formcontrolname="value"]';
+    expect(await tests.app.client.getValue(ruleTargetSelectSelector)).to.equal('body');
+    expect(await tests.app.client.getValue(ruleModifierInputSelector)).to.equal('test');
+    expect(await tests.app.client.getValue(ruleValueInputSelector)).to.equal('test');
```
255kb, line 104: Please consider creating a small helper for this

src/app/app.component.ts
izerozlu: I've moved the logic into EnvironmentService#duplicateRouteResponse.

src/app/stores/reducer.ts
izerozlu: The reason for creating a new UUID here was to be able to insert the duplicated route response just beneath the active route response. I didn't want to create another reducer which mostly looks like the `ActionTypes.ADD_ROUTE_RESPONSE` reducer.

I can't find a way to keep the logic in `ActionTypes.ADD_ROUTE_RESPONSE` reducer and not modifying the `action.routeResponse`.

But it seems to me creating a new reducer is the best and cleaniest way to go for this change.

What is your opinion? Is creating a new reducer overkill? Or maybe is it too much look-alike code?

255kb: I think that a new action is overkill when it's basically the same than ADD_ROUTE_RESPONSE. 
But you can modify this action, like it has been done for the ADD_ROUTE action. You will see that for this one the parameter afterUUID has been added in order to insert the duplicated route after its source. 
I think for duplicating route responses it should be done the same way. Not because it's the best one but so various implementations of similar features are the same.

izerozlu:
Alright, modifying the action is the way to go then. I think that's more
suitable as well.

izerozlu:
I've run Prettier formatting via `npx prettier` with the projects version which is `2.0.2`. It has changed over 90 files which are mostly not related to my work.
I don't suppose I should commit these changes. 
Should format the only files I've touched so far?

izerozlu:
I've resolved linting errors by the way. Only issue remaining is the formatting.

255kb:
> I've run Prettier formatting via `npx prettier` with the projects version which is `2.0.2`. It has changed over 90 files which are mostly not related to my work.
> I don't suppose I should commit these changes.
> Should format the only files I've touched so far?

Yes please, only format the files you touched, also if you can could you use vscode organize imports option? If you don't use vscode it's ok don't do it, but like this it's more compact, and vscode also get rid of unused imports, etc (that's maybe why you got 90 modified files)

izerozlu:
I've added commits with requested changes. 

Could you check those out when you have the time?

255kb:
> I've added commits with requested changes.
> 
> Could you check those out when you have the time?

Yes I saw, thank you :) 
I will try to review this as soon as possible

src/app/stores/actions.ts
```diff
@@ -289,11 +305,17 @@ export function setActiveRouteResponseAction(routeResponseUUID: string) {
  * Add a new route response
  *
  * @param routeReponse - route response to add
+ * @param relativeIndexInList - (optional) indicates where to put the new
+ * route response with relativity to active route response.
  */
-export function addRouteResponseAction(routeReponse: RouteResponse) {
+export function addRouteResponseAction(
+  routeReponse: RouteResponse,
+  relativeIndexInList?: number
```
255kb, line 313: I realize now that you can get rid of this parameters. Unlike a route which can be duplicated when active or not (through the context menu), a route response can only be duplicated when active. 
So the reducer code is enough and there is no need to pass this parameter. Sorry that I made you lose time.

src/app/stores/reducer.ts
```diff
@@ -739,10 +749,25 @@ export function environmentReducer(
               ...environment,
               routes: environment.routes.map((route) => {
                 if (route.uuid === state.activeRouteUUID) {
-                  return {
-                    ...route,
-                    responses: [...route.responses, newRouteResponse]
-                  };
+                  const responses = [...route.responses];
+                  if (![undefined, null].includes(action.relativeIndexInList)) {
```
255kb, line 753: As per my  previous comment this can be removed.

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { Tests } from 'test/lib/tests';
+import { expect } from 'chai';
+
+const tests = new Tests('basic-data');
+
+describe.only('Duplicate a route response', async () => {
```
255kb, line 6: Please remove the forgotten `.only`. And the `async`. Describes shouldn't be async.

Also, I think you could definitely "regroup" the tests. Duplicate once, in a first `it`, and check the headers, body, rules, etc in subsequent `it`. I usually have dependent `it` test cases, for each steps of the test, and use different `describe` to separate each test cases (with a call to `new Tests()` inside each describe to start with a fresh app/state). You can still keep it like this if you want. 👍

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    await tests.helpers.removeRouteResponse();
+    await tests.app.client
+      .element(routeResponseLabelInputSelector)
+      .setValue(label);
+    await tests.helpers.duplicateRouteResponse();
+
+    const labelText = await tests.app.client.getValue(
+      routeResponseLabelInputSelector
+    );
+    expect(labelText).to.equal(labelToCompare);
```
255kb, line 31: Please use the `assertElementValue` helper you created

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    await tests.helpers.removeRouteResponse();
+    await tests.app.client
+      .element(routeResponseStatusCodeSelectSelector)
+      .selectByValue(statusCode);
+    await tests.helpers.duplicateRouteResponse();
+
+    const statusCodeValue = await tests.app.client.getValue(
+      routeResponseStatusCodeSelectSelector
+    );
+    expect(statusCodeValue).to.equal(statusCode);
```
255kb, line 49: Please use the `assertElementValue` helper you created

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      '#route-response-headers .row.headers-list:nth-child(2) .form-control:first-child':
+        'test',
+      '#route-response-headers .row.headers-list:nth-child(2) .form-control:nth-child(2)':
+        'test'
+    };
+
+    Object.keys(selectorAndValueAssertionPairs).forEach((selector: string) => {
+      const valueToCompare = selectorAndValueAssertionPairs[selector];
+      tests.helpers.assertElementValue(selector, valueToCompare);
+    });
```
255kb, line 100: This code cannot work. If you test with any wrong value you will see that the test is still passing. There is no await on the helper on line 99. But you cannot just put an async on the `forEach` callback. What you need to do is to put a complete `it('', async () =>{])` inside the forEach. 
You can also see a warning saying "Unhandled promise rejection warning" which means that the test failed, but the failure wasn't caught (because the test did not wait). 

It's also fine for me if you get rid of the `forEach` and just call the helper four times. As long as it's clean and readable it does not need to be that DRY.

test/suites/route-response-duplicate.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    const selectorAndValueAssertionPairs = {
+      'app-route-response-rules .row select[formcontrolname="target"]': 'body',
+      'app-route-response-rules .row input[formcontrolname="modifier"]': 'test',
+      'app-route-response-rules .row input[formcontrolname="value"]': 'test'
+    };
+
+    Object.keys(selectorAndValueAssertionPairs).forEach((selector: string) => {
+      const valueToCompare = selectorAndValueAssertionPairs[selector];
+      tests.helpers.assertElementValue(selector, valueToCompare);
+    });
```
255kb, line 132: This code cannot work. If you test with any wrong value you will see that the test is still passing. There is no await on the helper on line 99. But you cannot just put an async on the `forEach` callback. What you need to do is to put a complete `it('', async () =>{])` inside the forEach. 
You can also see a warning saying "Unhandled promise rejection warning" which means that the test failed, but the failure wasn't caught (because the test did not wait).

src/app/stores/actions.ts
izerozlu: I need a parameter to indicate whether this action is a addition or duplication. If it's an addition, new route response is pushed, on the latter it's inserted relatively with the active route response.

I suppose I could not understand your statement with this comment. 

If I were to remove the routeReponse I lose the cloned/freshly created route response.
If I were to remove I lose the duplication indicator for this action.

Are you suggesting to remove both of them?

test/suites/route-response-duplicate.spec.ts
izerozlu: I suppose I'll go with this approach of grouping. Thanks for advice though.

test/suites/route-response-duplicate.spec.ts
izerozlu: Forgetting that `await` is bad on my side :) Instead of calling it four times or declaring `it` blocks with each `forEach` I've used `for...of` here. Did the work. Is that okay?

test/suites/route-response-duplicate.spec.ts
izerozlu: Forgetting that `await` is bad on my side :) Instead of calling it four times or declaring `it` blocks with each `forEach` I've used `for...of` here. Did the work. Is that okay?

src/app/stores/actions.ts
255kb: Sorry, you are right, I was probably tired when I wrote this. However, the `relativeIndexInList` is always set to '1' in the `duplicateRouteResponse` method in environment service. So you could either use a boolean instead, and compute the insertion index as you do in the reducer, or compute the index in the service directly.

test/suites/route-response-duplicate.spec.ts
255kb: I suppose it's better like you did as the assertions are all part of the same test.

izerozlu:
I've made commited some changes. Let me know if I've forgotten any of the problems.

255kb:
Perfect 👍 
Could you please fix the tests that are not passing? It has something to do with the route duplication button that is not displayed anymore.

255kb:
I fixed the test, it was the result of changes I made on master. It's ready to be merged. Thank you for your contribution!

izerozlu:
It's feels good to see the PR get approved :). Thanks for your guidance on it!

PR #328, opened 2020-08-16
Add null check for swagger v2 host field when importing

### LINKED ISSUE #326: Error when importing swagger v2 file
**Describe the bug**
When importing following swagger v2 file JS throw an error "cannot read property 'split' of undefined"

**To Reproduce**

```json
{
    "swagger": "2.0",
    "info": {
        "title": "os-admin-actions",
        "version": "Unknown"
    },
    "consumes": [
        "application/json"
    ],
    "produces": [
        "application/json"
    ],
    "paths": {
        "/v2/{tenant_id}/servers/{server_id}/action": {
            "parameters": [
                {
                    "name": "tenant_id",
                    "required": true,
                    "in": "path",
                    "type": "string",
                    "description": "The ID for the tenant or account in a multi-tenancy cloud.\n"
                },
                {
                    "name": "server_id",
                    "required": true,
                    "in": "path",
                    "type": "string",
                    "description": "The UUID for the server of interest to you.\n"
                }
            ],
            "post": {
                "operationId": "addFloatingIP",
                "summary": "Add floating IP address",
                "description": "Adds a floating IP address to an instance.\n",
                "produces": [],
                "responses": {
                    "200": {
                        "description": "200 response"
                    }
                }
            }
        }
    }
}
```

**Expected behavior**

cannot read property 'split' of undefined

**Screenshots**

截屏2020-08-05 上午11 21 38

**Mockoon version:**
v1.9.0

**OS / OS version:**
Mac OS Catalina 10.15.5

Spissable:
**Description**

Added a null check for the host field when importing a swagger v2 file. As the types correctly indicate, the host field is optional, it wasn't treated this way however.

Two side notes:
- Optional chaining would have been preferred, however the current angular compiler version doesn't allow a recent enough TypeScript version, which brings the feature (>= 3.7).
- These sort of bugs are preventable with the TypeScript compiler option `strict: true`. Not sure if it's left out intentionally, however when enabling it locally, it showed up quite some more type errors.

**Related Issue**

https://github.com/mockoon/mockoon/issues/326

**How Has This Been Tested?**
Import the swagger v2 file (or any other one without the `host` field) from the linked issue: it should work using the default port.

**Closing issues**

closes #326

### DIFF
--- a/src/app/services/openapi-converter.service.ts
+++ b/src/app/services/openapi-converter.service.ts
@@ -100,7 +100,7 @@ export class OpenAPIConverterService {
     );
 
     // parse the port
-    newEnvironment.port =
+    newEnvironment.port = parsedAPI.host &&
       parseInt(parsedAPI.host.split(':')[1], 10) || newEnvironment.port;
 
     if (parsedAPI.basePath) {


255kb:
Thanks for the contribution @Spissable !
Concerning, your two remarks it's true that it would have been caught with the strict flag, but it wasn't enabled since the beginning so it would need a little bit of work! It's on my list 👍 
And for the Typescript version we are currently constrained by Angular version 8. I plan to update soon but I need to first test all libraries that depends on it. Some of them may not have version for Angular > 9 (with the new Ivy rendering engine)

PR #340, opened 2020-10-03
 Remove api prefix when auto-mocking #333

### LINKED ISSUE #333: Remove API prefix when auto-mocking
When auto mocking, the newly created route contains the API prefix.
If a call is made to `/api/non-existing-route`, the route will be created with `/api/non-existing-route` as path, even if `/api` is already defined as the prefix.

Detect that a prefix is already set and create the route without it.

sleepypikachu:
**Description**
Check the prefix of the active environment when creating the route from the
environment logs. If the prefix is present in the request path then remove it
when creating the route.

**How Has This Been Tested?**

Manual testing on my local machine to verify behaviour, automated tests updated to cover the functionality.

Modify the environment-logs test spec to add a prefix and verify that the auto
mock removes the prefix in the created route.

**Closing issues**

Closes #333

### DIFF
--- a/src/app/services/environments.service.ts
+++ b/src/app/services/environments.service.ts
@@ -437,10 +437,16 @@ export class EnvironmentsService {
         routeResponse = this.schemasBuilderService.buildRouteResponse();
       }
 
+      const prefix = this.store.getActiveEnvironment().endpointPrefix;
+      let endpoint = log.url.slice(1); // Remove the initial slash '/'
+      if (prefix.length && endpoint.startsWith(prefix)) {
+        endpoint = endpoint.slice(prefix.length + 1); // Remove the prefix and the slash
+      }
+
       const newRoute: Route = {
         ...this.schemasBuilderService.buildRoute(),
         method: log.method.toLowerCase() as Method,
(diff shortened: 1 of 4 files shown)


src/app/services/environments.service.ts
```diff
@@ -437,10 +437,16 @@ export class EnvironmentsService {
         routeResponse = this.schemasBuilderService.buildRouteResponse();
       }
 
+      const prefix = this.store.getActiveEnvironment().endpointPrefix;
+      let endpoint = log.url.slice(1); // Remove the initial slash '/'
+      if (prefix.length && endpoint.startsWith(prefix)) {
```
255kb, line 442: I think it would be safer to test `prefix` directly instead of `prefix.length` in case it's null or undefined (even if it shouldn't) But no need to worry, I will do this small modification, and also check why the tests are not running on macOS.

255kb:
Thank you for your contribution! I will check why the tests are not running properly on macOS (I just noticed the same problem on Windows sometimes).

255kb (approved):

PR #343, opened 2020-10-06
Generate random ObjectId for mocking mongodb

### LINKED ISSUE #342: MongoDB's ObjectId support on templating
When using MongoDB, ObjectIds are used as Id, I use them in an API and this templating feature will save me a lot of time.

I added a template function to generate random [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/#description), they look like: 507f1f77bcf86cd799439011

https://github.com/mockoon/mockoon/compare/master...lukefx:templating_objectid

There is a GUID and faker UUID function but they are not exactly the same.

**Additional context**
I already created the function, but I need some time to understand how you text the templating, right now for me tests fail because the window takes a lot of time to start and load the test environment json file.

lukefx:
**Description**

{{objectId}} will generate a random MongoDB's ObjectId in the format of: 5f7c1ee149354fbcb2f9189b

**Related Issue**

Closes #342 

**Motivation and Context**

We use ObjectId as id and we expose them on our REST APIs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -649,6 +649,7 @@
           "dev": true,
           "optional": true,
           "requires": {
+            "bindings": "^1.5.0",
             "nan": "^2.12.1",
             "node-pre-gyp": "*"
           },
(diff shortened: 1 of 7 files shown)


lukefx:
@255kb I updated the code and added some tests, I also switched to a library so we have reproducible ObjectId based on DateTime or strings.
Tell me if is enough or needs more tests :)

255kb (approved):

255kb:
Thank you for this contribution @lukefx it's perfect! 
It will be in the next release but I cannot give any ETA for the moment. 
I will update the documentation on the website. 

I just noticed something after merging, the bson-objectid package can be moved to the dev dependencies (I did it). It's odd, but electron-builder actually use dependencies/devDependencies to define what needs to be installed when packaging the app. Thus, only some libraries end up in the dependencies as all the other one are being bundled by Webpack. 
It would be better if we had some sort of packageDependencies, prodDependencies and devDependencies ^^'

PR #383, opened 2020-12-06
Add route headers and rules count in tab

### LINKED ISSUE #299: Add route headers and rules count in tab
Without going to route response's headers and rules tabs it's not possible to know how much of them are setup. 

Add the counts `Rules (X)` in both tabs

TheDartCode:
**Parent issue** 

Closes #299

**Some details prior to the technical ones**
Since this is a small, contained feature, I took the initiative and implemented it without bugging you before doing the work.
However, if you don't like the implementation, I'll apply any change requests you make.

I also took the initiative to create the following behavior:
- When there are no headers, or no rules, the respective tab label does not contain any number indication. (So it is _Headers_ and _Rules_.)
- When one of the two tabs contains more-than-zero items, its label starts containing the count. (Examples: _Headers (1)_, _Rules (7)_ etc.)

- 2 methods are added in `src/app/app.component.ts` that calculate the suffixes for the tab labels. I wanted to follow the naming conventions that already exist there (e.g. `activeRouteResponseLastLog$`), so I ended up with these two names: `activeRouteResponseRulesLabelSuffix$` and `activeRouteResponseHeadersLabelSuffix$` (because we are talking about the suffix of the label -- all the aforementioned conditional logic is in there).
- I noticed that, in a fresh environment/route/response, the _Headers_ tab **always** contains an empty header, while the _Rules_ tab contains zero rules. This leads to an inconsistency where the _Headers_ tab always starts with the label _Headers (1)_, while the _Rules_ tab starts as plain _Rules_. This is a conceptual bug to me, as the _Headers_ tab is essentially empty. We either have to filter out empty headers while making the count, or start the tab with no header pre-created. Either way, my suggestion is that rules and headers should follow the same patterns. If one of them starts empty, then the other should do the same. (I'll happily implement both fixes if you'd like to).
- I created a `Headers && Rules tabs` section in the `ui-interactions.spec.ts` suite. During development, I was facing the issue that adding a rule did not update the respective tab label immediately, and I had to `pause` an arbitrary number of milliseconds in order for it to work smoothly. I am not experienced with angular or rxjs, so I suppose there is a better way to do it. Please advise and I will update the code accordingly.

**Checklist**

- [x] automated tests added

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -303,8 +303,8 @@
                         (click)="setActiveTab('HEADERS')">Headers</a>
                       </li>
                       <li class="nav-item">
-                        <a class="nav-link" [ngClass]="{'active': activeTab === 'RULES'}"
-                          (click)="setActiveTab('RULES')">Rules</a>
+                        <a class="nav-link" data-testid="rules-tab" [ngClass]="{'active': activeTab === 'RULES'}"
+                          (click)="setActiveTab('RULES')">Rules{{activeRouteResponseRulesLabelSuffix$ | async}}</a>
                       </li>
                       <li class="nav-item">
                         <a class="nav-link" [ngClass]="{'active': activeTab === 'SETTINGS'}"
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -300,11 +300,11 @@
                       </li>
                       <li class="nav-item">
                         <a class="nav-link" data-testid="headers-tab" [ngClass]="{'active': activeTab === 'HEADERS'}"
-                        (click)="setActiveTab('HEADERS')">Headers{{activeRouteResponseHeadersLabelSuffix$ | async}}</a>
+                        (click)="setActiveTab('HEADERS')">Headers{{(activeRouteResponseHeaderCount$ | async) > 0 ? ' (' + (activeRouteResponseHeaderCount$ | async)+')' : ''}}</a>
                       </li>
                       <li class="nav-item">
                         <a class="nav-link" data-testid="rules-tab" [ngClass]="{'active': activeTab === 'RULES'}"
-                          (click)="setActiveTab('RULES')">Rules{{activeRouteResponseRulesLabelSuffix$ | async}}</a>
+                          (click)="setActiveTab('RULES')">Rules{{(activeRouteResponseRuleCount$ | async) > 0 ? ' (' + (activeRouteResponseRuleCount$ | async)+')' : ''}}</a>
                       </li>
                       <li class="nav-item">
                         <a class="nav-link" [ngClass]="{'active': activeTab === 'SETTINGS'}"
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -299,11 +299,11 @@
                         (click)="setActiveTab('RESPONSE')">Status & Body</a>
                       </li>
                       <li class="nav-item">
-                        <a class="nav-link" data-testid="headers-tab" [ngClass]="{'active': activeTab === 'HEADERS'}"
+                        <a class="nav-link" [ngClass]="{'active': activeTab === 'HEADERS'}"
                         (click)="setActiveTab('HEADERS')">Headers{{(activeRouteResponseHeaderCount$ | async) > 0 ? ' (' + (activeRouteResponseHeaderCount$ | async)+')' : ''}}</a>
                       </li>
                       <li class="nav-item">
-                        <a class="nav-link" data-testid="rules-tab" [ngClass]="{'active': activeTab === 'RULES'}"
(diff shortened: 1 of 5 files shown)


255kb:
Thank you @TheDartCode for this contribution. 
To answer your questions: 
- The implementation looks good. Personally, I would have put the suffix ternary directly in the template, but I think it's rather a matter of taste.
- Concerning the counter, it's true that it's not very logical to always have a default empty header. But I think the counter should count what is in the UI and not only the valid headers. Otherwise, we could filter empty ones, but also the one without a value, or with invalid characters. I would prefer that you try to remove the default header in the  schema builder service https://github.com/mockoon/mockoon/blob/master/src/app/services/schemas-builder.service.ts#L28. It shouldn't have any serious consequence (or maybe a test or two will break).
- Tests. Thank you for adding them, this is really appreciated 😃 . I tested without the delays and it worked well locally (Windows), but it may be unavoidable in other systems. If you had to use them leave them. It may be possible that the update occurs during another change detection cycle, and the testing framework has to wait for the UI to be updated. I try to avoid adding artificial delays but sometimes it cannot work without. One thing about the `data-testid` attributes, it's clean but we don't use this right now, so to keep it consistent with the rest of the tests, I would prefer to use a class name or `:nth-of-type` instead, if you agree.

TheDartCode:
Thanks for your time!
I will do the following changes, and then get back to you:
- Move the ternary inside the template.
- Remove the default header, as per your guidelines.
- Remove `data-testid` attributes, use `:nth-of-type` selector (I agree with consistency as well).
- Try issuing the PR without the artificial delays and see how it behaves in CI.

Cheers!

src/app/app.component.html
```diff
@@ -300,11 +300,11 @@
                       </li>
                       <li class="nav-item">
                         <a class="nav-link" [ngClass]="{'active': activeTab === 'HEADERS'}"
-                        (click)="setActiveTab('HEADERS')">Headers</a>
+                        (click)="setActiveTab('HEADERS')">Headers{{(activeRouteResponseHeaderCount$ | async) > 0 ? ' (' + (activeRouteResponseHeaderCount$ | async)+')' : ''}}</a>
```
255kb, line 303: Could you please subscribe only once by using an `*ngIf` and `as`:
```
<a *ngIf="activeRouteResponseHeaderCount$ | async as activeRouteResponseHeaderCount" ....>
```

255kb (changes requested):
@TheDartCode Thank you for the changes 👍 
It looks like the tests are working without the `pause()`. 

I just requested two small changes: in the template you can avoid subscribing twice with the async pipe by using an ngIf and assigning the async pipe result to a variable.

src/app/app.component.html
TheDartCode: Thanks for the tip! TIL 😀

TheDartCode:
> @TheDartCode Thank you for the changes 👍
> It looks like the tests are working without the `pause()`.

@255kb I think you are mistaken 🙂 The CI checks have failed, and if you see the error message it's because it was expecting to find an updated count (`Rules (1)`) but instead it found the previous (`Rules`). If you see the commit history above, you can see that the last commit (removing the artificial delays) broke the process. The immediately previous one had a successful CI run. What would you like me to do?

255kb:
@TheDartCode OK I misunderstood then 😄 In this case keep the delays!

TheDartCode:
Requested changes done, and the delays have (sadly) been reverted! Checks pass!

255kb:
Thank you @TheDartCode !
After all, test are done by a robot clicking in a UI, so we can expect some uncertainty 😄 

I don't see the async pipe change, did you push it? 🤔

src/app/app.component.html
TheDartCode: Whoops, totally forgot about it!

I tried to implement it just now, but as far as I can see the `*ngIf` hides the `<a />` when rule/header count equals 0.
If I create 2 separate anchor tags, one for when rules exist and one for when they don't it will end up with two subscriptions again.
Can you think of a workaround to that?

255kb: I think you could use `ng-container` like this:
```
<a>
Header<ng-container *ngIf="obs$ | async as obs"> ({{obs}})</ng-container>
</a>
```

TheDartCode:
@255kb Everything should be OK now, please have another look 🙂

255kb (approved):
@TheDartCode it's perfect! Thank you for your contribution 🚀 
It will be in the next release, but I cannot give any ETA as we just started.

TheDartCode:
Great! Thanks again, I'll look for another opportunity for contribution in Mockoon now 👀 
No hurry regarding the release 😁

PR #394, opened 2021-01-03
Move(duplicate) a route to another environment

### LINKED ISSUE #308: Support for moving an api from one environment to another environment
I can't drag or move an api from one environment to another environment.

Support for dragging api item.

Add a menu to api item for moving api item to another environment.

**Additional context**
Best API mock util I've ever used, good work.

izerozlu:
**Parent issue** 
Closes #308

I've added a new context menu item in route context menu. With this new item the user opens a new dialog to see the summary of the desired route to move and selects the target environment.

For modal controls, store is used. Actions and reducers decides whether to open or close the dialog.

When the user chooses which environment to move the route to, route duplicates in the target environment and application focuses on the target environment and the freshly duplicated route.

For the duplication, in hand actions used.

**Checklist**

- [x] automated tests added


255kb:
Thank you @izerozlu for your contribution. I've been quite busy the past weeks and wasn't able to review yet. I will do it as soon as possible!

255kb:
I started reviewing quickly and I wonder if it would make sense to rename all the methods, actions, etc. from `moveXXX` to `duplicateToEnvironmentXXXX` as we are in fact duplicating and not moving the route. I find it a little bit ocnfusing. What do you think @izerozlu?

izerozlu:
I had that kind of thoughts while naming those methods as well. I felt the need to seperate this "movement" process from other kinds of processes so i tried have another scope for these. But as a matter of fact I kind of have the same feeling with you so renaming those methods would actually be better.

I will look into that asap.

255kb:
I tested your feature, and it's working well congrats 😃 
I just found a small bug. When you update the state of the target env you set the status `running` at true, but if the env is not running it puts it in an odd state, and inconsistent buttons colors. it need to be checked first.
Also, I would like to change and simplify the design of the modal a bit so it matches the rest of the look and feel: no rounded corners, and lighter borders, each entry taking the modal's full width, etc. I also think the path and methods at the top could be removed and replaced by the title, and showing the environment name only should be enough, what do you think? (I really want to avoid overloading the UI with unneeded information, which is definitely not easy!)
I have some time in the coming days, so I can easily jump on this and do the modification if you are ok. Let me know.

izerozlu:
@255kb Hey,

First of all, I will look into bug you've mentioned.

I'm okay with the design changes, actually I've tried to come up with a dialog which matches with the design in place but of course you would know more than me about this :D

About the last suggestion I think leaving the indicators of what is being moved/duplicated is crucial. Maybe user misclicked a route or maybe they didn't notice they've chosen the wrong method etc. I really o think showing the route related information is important but the last call is yours of course.

izerozlu:
PS: I'll look into method name changes as well. I've been quite busy with the work and family lately.

255kb:
@izerozlu No problem, I will let you fix the bug (you will see the same code for assessing if a restart is needed in the reducer, just copy it, don't bother DRYing the reducer, I need to do it one day 😅) and do the renaming.
And if you agree I will do the design changes after, but you are right, we can keep the method+path as a reminder, just not sure putting it at the top of the modal if the best place. I would prefer the title, then the reminder of what you are moving, then the list of env (simplified, with only names). 
Let me know if you need anything! Thanks

izerozlu:
@255kb I've implemented the design change, bug fix, and name changes. Please check out when you have the time.

255kb:
Thanks @izerozlu it works great! 
I made some slight changes (locally) to the modal design: 
image
Let me know if you like it. 
I will finalize the design and merge this evening so it can make it for the next release.

izerozlu:
You know your design better than me, so it's okay for me.

Go for it!

10 Şub 2021 Çar ÖS 6:40 tarihinde Guillaume <notifications@github.com> şunu
yazdı:

255kb (approved):

255kb:
Great! Thanks again for your contribution 👍

PR #464, opened 2021-04-29
Allow for custom status codes

### LINKED ISSUE #199: Ability to add Custom HTTP status codes
Currently, there is a dropdown with a list of pre-defined HTTP status codes. It would be very helpful if the user could add custom HTTP status codes.

255kb:
- create a new custom select component using bootstrap dropdown, with filter capability, keyboard navigation
- Group statuses by categories
- Add non official status codes (Cloudflare, nginx, etc)
- redo the HTTP methods select with the new component and improve the colors
Closes #199

**Parent issue** 

Closes #199 

**Checklist**

- [x] automated tests added


255kb:
After testing a bit, I think it's working well. 
However, if you create a first route with no rules (let's say a 200), and a second one with `request_number=1`, you can never get the first route. 
I wonder if this is a logical behavior. Until now, the rules system was serving the first response by default if no rule matched. I have the feeling that changing this default behavior could make things a little bit confuse. What do you think? 

Also, I don't really like removing inputs from the UI. Then things are not aligned and it looks a little bit weird in my opinion. I would rather see something like a disabled input: 
image
But CSS is not ready for this. I think when everything is done and tested, I will work on the UI part if it's ok for you.

fabhoarau (approved):

PR #476, opened 2021-05-12
closes mockoon/mockooon#237

### LINKED ISSUE #237: Add option to prevent API prefix to be forwarded to proxied server
When setting up an API prefix, it is forwarded as is to the proxied route. 
It can make sense that requests are proxied without the API prefix. 

Example: 
1. `localhost:3000/myprefix` is declared in Mockoon, with no route, and proxy mode active (set to `http://localhost: 3010/`)
2. when calling `localhost:3000/myprefix/foo` it is forwarded to `http://localhost: 3010/myprefix/foo`

It could be useful to have it forwarded to `http://localhost: 3010/foo`, ignoring the prefix declared in Mockoon. 
This should be store in the settings.

barefootdeveloper:
Closes #237 

I added:
* A ui option on Environment Settings to allow for removing the prefix from forwarded requests.
* Added a test suite that tests the removal of the prefix.


255kb:
Thanks for the contribution @barefootdeveloper I will review a little bit later this week.

barefootdeveloper:
You’re welcome. Thanks for a great tool!  I didn’t bump the common packages versions and reference the new version in the dependent projects. I think that’s what is breaking the build but I know maintainers often like to do that. If you want anything changed I’m happy to do it.

255kb:
You're welcome :) Thanks for using it and taking time to contribute!
Don't worry, I need to test everything together first and create releases so it gets published on NPM.

test/suites/migrations.spec.ts
```diff
@@ -126,4 +126,20 @@ describe('Environments migrations', () => {
...
+    const tests = new Tests(filesPath);
+
+    it('Should add "proxyRemovePrefix" properties to environments', async () => {
+      await tests.helpers.waitForAutosave();
+
+      await tests.helpers.verifyObjectPropertyInFile(
+        './tmp/storage/environments.json',
+        ['0.proxyRemovePrefix'],
+        false,
+        true
```
255kb, line 141: I think you can remove the last param `true` as we want to test if it's `false`.

255kb (approved):
Thanks @barefootdeveloper it's perfect! 
One very small modification, but I can do it. I also need to migrate all the test data samples (there is a script for that). 
I will integrate this change for the next release. 
Thanks again for your work 😃

test/suites/migrations.spec.ts
barefootdeveloper: I interpreted those parameters differently, but digging into that function I agree it should be pulled out. I updated the branch.

255kb (approved):

255kb:
Thanks for the changes @barefootdeveloper 
I will merge and release next week.

PR #484, opened 2021-05-20
Closes #479

### LINKED ISSUE #479: Decode gzip/deflate automatically when mocking a proxied request
When the origin server that I'm proxying responses with gzip/deflate encoded body, mockoon just shows the raw content when mocking the request. This is annoying since it's impossible to quickly adjust the response in the mock. See https://github.com/mockoon/mockoon/discussions/453

- mockoon should automatically decode these response bodies when creating a mock
- the returned mocked response should then be encoded again
- an icon could indicate that the original response was and and the mocked one will be gzipped
- optionally there could be a way of switching back to the raw content, but I personally don't have a use case for that

- there also could be a checkbox in the proxy settings to disable this feature, but I would prefer the icon/toggle described above

**Additional context**

image

I worked around it by setting a proxy request header, but that's not very intuitiv:
image

rubaiate:
createRouteFromLog: update encoding headers
  Indicate response content encoding in ui

**Parent issue** 

Closes #479

**Checklist**

- [ ] automated tests added

### DIFF
--- a/src/renderer/app/services/environments.service.ts
+++ b/src/renderer/app/services/environments.service.ts
@@ -474,10 +474,25 @@ export class EnvironmentsService {
       if (log.response) {
         const headers: Header[] = [];
         log.response.headers.forEach((header) => {
+          if (
+            header.key === 'content-encoding' ||
+            header.key === 'content-length'
+          )
+            return;
+
           headers.push(
             this.schemasBuilderService.buildHeader(header.key, header.value)
           );
         });
+
+        if (log.response.body) {
+          headers.push(
+            this.schemasBuilderService.buildHeader(
+              'content-length',
+              log.response.body.length.toString()
+            )
+          );
+        }
 
         routeResponse = {
           ...this.schemasBuilderService.buildRouteResponse(),
--- a/src/renderer/app/components/environment-logs.component.html
+++ b/src/renderer/app/components/environment-logs.component.html
@@ -383,7 +383,11 @@
                     class="material-icons"
                     >unfold_more</i
                   >
-                  Body ({{ !selectedLog.response?.body ? 'none' : 'raw' }})
+                  Body ({{
+                    !selectedLog.response?.bodyState
+                      ? 'none'
+                      : selectedLog.response?.bodyState
+                  }})
                 </div>
                 <div *ngIf="!selectedLog.response.binaryBody" class="ml-auto">
                   <span
--- a/src/renderer/app/components/environment-logs.component.ts
+++ b/src/renderer/app/components/environment-logs.component.ts
@@ -87,6 +87,20 @@ export class EnvironmentLogsComponent implements OnInit {
           if (environmentLog.response.body) {
             environmentLog.response.truncatedBody =
               this.dataService.truncateBody(environmentLog.response.body);
+
+            const contentEncoding = environmentLog.response.headers.find(
+              (header) => header.key.toLowerCase() === 'content-encoding'
+            )?.value;
+
+            if (
+              contentEncoding === 'gzip' ||
+              contentEncoding === 'br' ||
+              contentEncoding === 'deflate'
+            ) {
+              environmentLog.response.bodyState = 'unzipped';
+            } else {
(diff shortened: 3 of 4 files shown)


255kb:
Thanks for opening the PR @rubaiate 
Also, you don't need to close a PR if you want to make changes, you can push a new commit on your branch and the PR will be updated.

255kb:
I will probably review this in two weeks. Next week I will have limited access to my computer, so don't worry if you don't hear from me in the coming days :)

barefootdeveloper:
This is going to be a really significant feature, thank you for taking it on! 

I have one concern about the current implementation. I believe if we create a mocked response off of proxied request with a compressed response, it would still create an invalid response. This is because the headers stay, showing that the response is compressed and encoded.

The alternatives I see are:

1) When decoding the response, alter the headers as well so it becomes a valid response.
2) Move the decoding to the UI so that you can see the decoded response. Add support for compressing responses.

The first option is easier, but results in a different endpoint behavior from the original. I believe this is the approach that @255kb suggested.

The second option seems the most ideal since it maintains the original API behavior, but I suspect most callers of an API would handle this transparently so it might be mostly an academic argument.

src/renderer/app/services/environments.service.ts
```diff
@@ -474,11 +474,26 @@ export class EnvironmentsService {
       if (log.response) {
         const headers: Header[] = [];
         log.response.headers.forEach((header) => {
+          if (
+            header.key === 'content-encoding' ||
```
barefootdeveloper, line 478: I think we may need transfer-encoding added to this list to facilitate the mock this endpoint.

rubaiate: As of now, transfer-encoding is not checked also while decoding messages. Maybe need to check at both decoding and header update. I'll check what is possible.
Thanks.

255kb:
> This is going to be a really significant feature, thank you for taking it on!
> 
> I have one concern about the current implementation. I believe if we create a mocked response off of proxied request with a compressed response, it would still create an invalid response. This is because the headers stay, showing that the response is compressed and encoded.
> 
> The alternatives I see are:
> 
> 1. When decoding the response, alter the headers as well so it becomes a valid response.
> 2. Move the decoding to the UI so that you can see the decoded response. Add support for compressing responses.
> 
> The first option is easier, but results in a different endpoint behavior from the original. I believe this is the approach that @255kb suggested.
> 
> The second option seems the most ideal since it maintains the original API behavior, but I suspect most callers of an API would handle this transparently so it might be mostly an academic argument.

Indeed, I supported the version 1, where you can create a mock from compressed content while still being able to modify it. Also, I don't see the automatic mock creation as a way to recreate 1:1 what has been received by mockoon but rather to have a base to create a new mock. 
If we want to support encoding of outgoing responses, it should be an activable option at the route or env level, and part of another feature request.

255kb:
> This is going to be a really significant feature, thank you for taking it on!
> 
> I have one concern about the current implementation. I believe if we create a mocked response off of proxied request with a compressed response, it would still create an invalid response. This is because the headers stay, showing that the response is compressed and encoded.
> 
> The alternatives I see are:
> 
> 1. When decoding the response, alter the headers as well so it becomes a valid response.
> 2. Move the decoding to the UI so that you can see the decoded response. Add support for compressing responses.
> 
> The first option is easier, but results in a different endpoint behavior from the original. I believe this is the approach that @255kb suggested.
> 
> The second option seems the most ideal since it maintains the original API behavior, but I suspect most callers of an API would handle this transparently so it might be mostly an academic argument.

Indeed, I supported the version 1, where you can create a mock from compressed content while still being able to modify it. Also, I don't see the automatic mock creation as a way to recreate 1:1 what has been received by mockoon but rather to have a base to create a new mock. 
If we want to support encoding of outgoing responses, it should be an activable option at the route or env level, and part of another feature request.

255kb:
@rubaiate Let me know if you want me to take it from here, finalize and add tests.

rubaiate:
Great. 👌

PR #486, opened 2021-05-23
Closes mockoon/mockoon#469

### LINKED ISSUE #469: [Enhancement] Add ability to change host configuration for API hosts.
Not a problem, but being able to restrict API availability to localhost would be nice.

It would be useful to allow the underlying host configuration for the mock APIs to be configured. This would allow a user to restrict what interfaces the mock server would respond to.

For the interface I was thinking of a drop-down picking between All Hosts (0.0.0.0) and local host only (127.0.0.1) but it might be useful to expand it later to allow for manual entry in scenarios where a machine has multiple network cards.

I believe this will require changes to:

- commons/src/models/environment.model.ts - Add Host Property
- commons-server/src/libs/server.ts - Make use of host property, default to 0.0.0.0 if not defined.
- A new component to be referenced from mockoon/src/renderer/app/app.component.html

I didn't dig into the CLI tool, it may need changes there to allow for configuration of the host.

I was actually thinking of picking up this work since I think it's a nice value-add, but not really necessary since you can simply configure firewall rules to restrict outside access.

barefootdeveloper:
Closes #469 
Changes HTTPS settings to Server Settings
Changes Enable checkbox to HTTPS
Adds Localhost Only checkbox under Server Settings
Makes the environments menu show the configured hostname instead of a static 0.0.0.0

The Localhost Only checkbox just determines if 127.0.0.1 or 0.0.0.0 gets set as the hostname configuration value.

### DIFF
--- a/src/renderer/app/app.component.html
+++ b/src/renderer/app/app.component.html
@@ -97,10 +97,9 @@
                 <span class="input-group-text"
                   ><i
                     class="material-icons"
-                    ngbTooltip="Server available on all network interfaces (localhost, 127.0.0.1, etc)"
+                    [ngbTooltip]="hostnameTooltip$ | async"
                     >power</i
-                  >
-                  0.0.0.0 :</span
+                  >&nbsp; {{ activeEnvironment.hostname }} :</span
                 >
               </div>
               <input
@@ -380,14 +379,14 @@
                 <!-- flexbox container -->
                 <div>
                   <app-title-separator
-                    heading="HTTPS"
-                    icon="https"
-                    iconClasses="text-warning"
+                    heading="Server Settings"
+                    icon="storage"
+                    iconClasses="text-primary"
                     docLink="https"
                   ></app-title-separator>
 
                   <div class="form-inline pl-4">
-                    <div class="custom-control custom-checkbox">
+                    <div class="custom-control custom-checkbox mr-4">
                       <input
                         type="checkbox"
                         class="custom-control-input"
@@ -395,7 +394,18 @@
                         formControlName="https"
                       />
                       <label class="custom-control-label" for="HTTPS"
-                        >Enable</label
+                        >HTTPS</label
+                      >
+                    </div>
+                    <div class="custom-control custom-checkbox">
+                      <input
+                        type="checkbox"
+                        class="custom-control-input"
+                        id="localhostOnly"
+                        formControlName="localhostOnly"
+                      />
+                      <label class="custom-control-label" for="localhostOnly"
+                        >Localhost Only</label
                       >
                     </div>
                   </div>
--- a/src/renderer/app/app.component.ts
+++ b/src/renderer/app/app.component.ts
@@ -119,6 +119,7 @@ export class AppComponent implements OnInit, AfterViewInit {
   public toasts$: Observable<Toast[]>;
   public focusableInputs = FocusableInputs;
   public statusCodeValidation = StatusCodeValidation;
+  public hostnameTooltip$: Observable<string>;
   private injectHeaders$ = new Subject<Header[]>();
   private logger = new Logger('[COMPONENT][APP]');
   private closingSubscription: Subscription;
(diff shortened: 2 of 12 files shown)


255kb (changes requested):
I think the implementation could be simplified a bit. `environment.hostname` is a string (either 127.0.0.1 or 0.0.0.0) but it could have been a boolean. 
Then, we could get rid of some code, like the conversion from `localhostOnly` to `environment.hostname` in the `valueChange` subscription.
I think it should be called `localhostOnly` everywhere and be used as a simple flag. 
Also, in commons-server, I would rather listen on `127.0.0.1` if `localhostOnly` is `true` or pass `null` to the `listen()` function if  it's `false`. It also reduces the possible mistakes when using the CLI and editing the data files manually. 

One more thing, I would like to see a test for this feature (in mockoon/mockoon), maybe by checking the available network interfaces with Nodejs `os.networkInterfaces()`. I am happy to add this test if you want, and make it compatible with the CI. Let me know.

barefootdeveloper:
> I think the implementation could be simplified a bit. `environment.hostname` is a string (either 127.0.0.1 or 0.0.0.0) but it could have been a boolean.

I thought about that when we settled on simplifying to localhost only, but I decided against it because we had also talked about the potential for the CLI (--adapter flag) to benefit from specific hostnames. I choose this approach because it allowed us to keep the UI simple but potentially expand it in the future without any server/model changes. It's also a closer representation of what the server itself is doing.

I'm happy to look into the unit test.

255kb:
> > I think the implementation could be simplified a bit. `environment.hostname` is a string (either 127.0.0.1 or 0.0.0.0) but it could have been a boolean.
> 
> I thought about that when we settled on simplifying to localhost only, but I decided against it because we had also talked about the potential for the CLI (--adapter flag) to benefit from specific hostnames. I choose this approach because it allowed us to keep the UI simple but potentially expand it in the future without any server/model changes. It's also a closer representation of what the server itself is doing. I'll also note that it makes some things easier like knowing what to output on the environments menu and the top menu bar without putting switch statements in those locations.

This makes perfect sense. Then keep the implementation as it is! 
One thing is still bothering me. Now that `hostname` is exposed to the user (through the data file) it may be modified manually. So we should account for two type of errors in [commons-server](https://github.com/mockoon/commons-server/blob/main/src/libs/server.ts#L78) and display them in a toast in the app (https://github.com/mockoon/mockoon/blob/f52eef8c6288304b10804298eeb167aa73d61f75/src/renderer/app/constants/messages.constants.ts#L44): 
- Unavailable IP (here a typo 19 -> 192): `Error: listen EADDRNOTAVAIL: address not available 19.168.1.1:3000`
- Erroneous hostname (here 'whatever'): `Error: getaddrinfo ENOTFOUND whatever`

Both cases should also be covered with tests.

barefootdeveloper:
While I was writing these unit tests the other thing I was thinking we could do that's sort of a middle ground is making the hostname optional. If no hostname is provided it will not pass one to listen and if one is provided it would. The localhost checkbox would work as it does today except it'd change from 0.0.0.0 to no value at all.

255kb:
> While I was writing these unit tests the other thing I was thinking we could do that's sort of a middle ground is making the hostname optional. If no hostname is provided it will not pass one to listen and if one is provided it would. The localhost checkbox would work as it does today except it'd change from 0.0.0.0 to no value at all.

I think that I prefer your first implementation. Having `0.0.0.0` always by default, and switching to `127.0.0.1` with the app checkbox. It gives enough flexibility when we implement this in the CLI and it's clear enough.

barefootdeveloper:
That works for me, it stops the server from ever answering on IPv6, not sure if that changes anything. I can't imagine too many folks use IPv6 for their locals, but they may on devices that run the cli.

We could use ::1 which I believe ends up working for both IPv4 and IPv6 but that'd require an IPv6 stack to be available, which I imagine it is most of the time.

I realized all of this whenever I wrote the unit test and it was cycling through IPv6 addresses.

255kb:
I see. Then, why not listen by default on `0.0.0.0` and on `localhost` if the checkbox is checked? After all, the checkbox label says "Localhost only". And I would say that most of the time localhost would resolve to `127.0.0.1` or `::1`. I don't think many people are binding their localhost to something else.

barefootdeveloper:
I think I have covered everything we talked about.

255kb (approved):

PR #488, opened 2021-05-27
Closes mockoon/mockoon#267

### LINKED ISSUE #267: Extend environment log to x entries, it is too small! Add option to export environment log to file for inspection.
I'm having an asynchronous process call back to the Mockoon service. Inspecting these messages is time consuming from the server logs on the service calling Mockoon. 

Ideally we would have a configurable environment log size so optionally a large number of entries is retained. i.e. default 60, max 1028 entries in the environment log.

I'm looking to have these exported to file for inspection at a later time and for test audit purposes as well. This is linked to the need to have a larger environment log. If the environment log was written to file by default, this requirement would be superseded by this. 

If only 60 entries were in the UI, but all entries were pushed out to file then there wouldn't be a need for a larger log in the app or an export function.

Thanks for your time. Great tool.

barefootdeveloper:
**Parent Issue**
Closes #267 

* Renamed MaxLogsPerEnvironment to DefaultMaxLogsPerEnvironment
* Added New Setting that Inherits DefaultMaxLogsPerEnvironment
* Modified the Reducer to use a while loop in case the max logs setting is reduced
* Added a unit test and a couple helpers to facilitate the test.
* Removed distinctUntilChanged() because in my testing it seems like the comparer is almost always false except for the initial screen draw and it does a lot of work whenever the log list is long. I'll admit I don't know the original intent of this change so if i'm overlooking something please tell me and I'm happy to put it back.

### DIFF
--- a/src/renderer/app/components/environment-logs.component.ts
+++ b/src/renderer/app/components/environment-logs.component.ts
@@ -63,8 +63,7 @@ export class EnvironmentLogsComponent implements OnInit {
 
   ngOnInit() {
     this.environmentLogs$ = this.store
-      .selectActiveEnvironmentLogs()
-      .pipe(distinctUntilChanged());
+      .selectActiveEnvironmentLogs();
 
     this.activeEnvironmentLogUUID$ = this.environmentLogs$.pipe(
       mergeMap(() => this.store.selectActiveEnvironmentLogUUID())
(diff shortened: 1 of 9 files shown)


barefootdeveloper:
I realize I broke a unit test in Environment logs by setting the environment log limit too low for one of the tests.  I'm fixing that  Not sure what to make of the headers test failing.

255kb:
Do not worry, the header test is randomly failing, not sure why and didn't have time to investigate yet. But it's not related to your PR.
I will review the PR in the coming days 👍 
One thing, for the next contribution, if you could comment on the original issue so I can assign it to you and put it on the roadmap. Also, it gives us the occasion to discuss a little bit the implementation first. For this one it was obvious and I would have chosen the same implementation, but the issue was exploring different possibilities. It's also a matter of documenting and explaining why a feature was done in one way and not another. Thank you 😃

PR #500, opened 2021-06-08
Feature/configure log count

### LINKED ISSUE #267: Extend environment log to x entries, it is too small! Add option to export environment log to file for inspection.

barefootdeveloper:
I renamed the branch to try to get it more-in-line with how you name your branches and it closed the PR, sorry about that.

Parent Issue
Closes #267

Technical implementation details
* Renamed MaxLogsPerEnvironment to DefaultMaxLogsPerEnvironment
* Added New Setting that Inherits DefaultMaxLogsPerEnvironment
* Modified the Reducer to use a while loop in case the max logs setting is reduced
* Added a unit test and a couple helpers to facilitate the test.
* Removed distinctUntilChanged() because in my testing it seems like the comparer is almost always false except for the initial screen draw and it does a lot of work whenever the log list is long. I'll admit I don't know the original intent of this change so if i'm overlooking something please tell me and I'm happy to put it back.

### DIFF
--- a/src/renderer/app/components/environment-logs.component.ts
+++ b/src/renderer/app/components/environment-logs.component.ts
@@ -63,8 +63,7 @@ export class EnvironmentLogsComponent implements OnInit {
 
   ngOnInit() {
     this.environmentLogs$ = this.store
-      .selectActiveEnvironmentLogs()
-      .pipe(distinctUntilChanged());
+      .selectActiveEnvironmentLogs();
 
     this.activeEnvironmentLogUUID$ = this.environmentLogs$.pipe(
       mergeMap(() => this.store.selectActiveEnvironmentLogUUID())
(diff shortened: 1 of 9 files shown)


255kb (approved):

255kb:
Thanks @barefootdeveloper this looks great! I just changed the wording in the settings modal a little bit.

PR #636, opened 2022-02-11
Feature/query string display

### LINKED ISSUE #635: Unhelpful preview for structured QueryParams in the request log
There'a s problem with the preview of query parameters in the logged request if they are structured. The preview shows `[object Object]` instead of the real contents of the params.

### Discussed in https://github.com/mockoon/mockoon/discussions/634

<div type='discussions-op-text'>

<sup>Originally posted by **marszysz** February  4, 2022</sup>
Hello,
It would be very helpful if one could see the raw query string in the request log. For now, if the query contains a structure, the preview isn't helpful at all:
obraz
Ideally, I would love to see some link like the one `View full body in editor` that would allow me to inspect the whole raw query string. For now, there's no means for doing this.</div>

timon:
**Parent issue**

Closes #635
Requires mockoon/commons#33 and mockoon/commons-server#40

This relies on server providing raw query string in `Transaction.request.query` and parsed query params in `Transaction.request.queryParams`.

Two methods added to data services to iterate over `queryParams` and transfrom them into `key: value` pairs, calculating key to be nested path in the `queryParams` object

**Checklist**

- [ ] automated tests added
- [ ] ~data migration added (if applicable)~
- [ ] ~data migration automated tests added (if applicable)~


timon:
Somehow after changes to `commons` and `commons-server` were rebased, this does not build anymore :(

255kb:
> Somehow after changes to `commons` and `commons-server` were rebased, this does not build anymore :(
Sorry, I had WIP on a branch that should have been merged before you start working.
It's merged now, you should be able to rebase on main and have a passing build.

timon:
Ok, turns out with typescript you can't just reference github repos for node modules.
So basically I'll need to wait for https://github.com/mockoon/commons/pull/33 and https://github.com/mockoon/commons-server/pull/40 to hit the main branch and to be released on npm.

I'll drop the package.json changes then.

255kb:
@timon This looks good! One thing thought, I think it would be more consistent to also hide the "Raw query string" when the query params section is collapsed, instead of keeping it shown. 
I published the two libs, you can update them on your branch. 
Also, could you make sure to format both TS and HTML code with Prettier? Thanks!

src/renderer/app/services/data.service.ts
```diff
@@ -85,6 +86,10 @@ export class DataService extends Logger {
     };
   }
 
+  public formatQueryParams(requestParams: Transaction['request']['queryParams']) : { name: string; value: string }[] {
+    return this.formatQueryParamsWithPrefix('', requestParams);
+  }
+
```
255kb, line 92: Do we really need this intermediate function? I guess you could use directly `this.formatQueryParamsWithPrefix('', requestParams)` on line 75
Also, if kept, I think this method should be private.

timon: I have introduced this mostly for readability reasons. I can make it go, if you like

255kb: Let's keep it then, but can you make it `private`?

255kb:
@timon I would like to finalize the current release. If you want I can take it from here, write the tests and merge. Let me know.

src/renderer/app/services/data.service.ts
timon: Done

timon:
> I think it would be more consistent to also hide the "Raw query string" when the query params section is collapsed

At some point I thought of using raw query string instead of parameters count on the section, 
like: `Query Params (a=b&c[d]=e...)`, but I didn't get to try that. However with long query strings it might turn out not be really helpful.

I have adjusted code on your comments. As I have no idea on tests currently, it would be great if you could take it on from this point

255kb:
Thank you @timon for your contribution!
I will merge and add the tests.

255kb (approved):

Issue #689, opened 2022-04-21
Uncaught exception raised by `decodeURI` results in a process crash


martin-bucinskas:
**Describe the bug**
The usage of `decodeURI` in `commons-server/src/libs/utils.ts` may result in an uncaught exception raised, crashing the process.

```
/my-project/node_modules/@mockoon/commons-server/dist/cjs/libs/utils.js:108
            query: requestUrl ? decodeURI(requestUrl.search.slice(1)) : null,
                                ^

URIError: URI malformed
    at decodeURI (<anonymous>)
    at CreateTransaction (/my-project/node_modules/@mockoon/commons-server/dist/cjs/libs/utils.js:108:33)
    at ServerResponse.<anonymous> (/my-project/node_modules/@mockoon/commons-server/dist/cjs/libs/server.js:112:81)
    at ServerResponse.emit (events.js:400:28)
    at emitCloseNT (_http_server.js:820:8)
    at processTicksAndRejections (internal/process/task_queues.js:81:21)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
...
npm ERR! Exit status 1
npm ERR! 
...
```

**To Reproduce**
Request sent with a query parameter containing a malformed URI.
See the HTTPie requests below.

```bash
http --follow --timeout 3600 GET 'localhost:3001/my-endpoint?test=%3C%%3D4242%20%%3E'
```

Or equivalent non encoded version

```bash
http --follow --timeout 3600 GET 'localhost:3001/my-endpoint?test=<%=4242 %>'
```

**Expected behavior**
The expected behaviour would be to handle the `decodeURI` exception in a try/catch statement and handle accordingly (either log as a warning or an error, and replace the malformed URI with either an empty string or a sanitised string.

**Screenshots**
N/A

**Mockoon version:**
1.18.1

**OS / OS version:**
N/A

PR #712, opened 2022-05-18
Fix decodeURI process crash

tsgodara:
Closes https://github.com/mockoon/mockoon/issues/689

Technical implementation details

Checklist

### DIFF
--- a/packages/commons-server/src/libs/utils.ts
+++ b/packages/commons-server/src/libs/utils.ts
@@ -111,9 +111,9 @@ export function CreateTransaction(
     let queryString = requestUrl.search.slice(1);
     try {
         queryString = decodeURI(queryString);
-    } catch (e) {
+    } catch (err) {
         queryString = "";
-        console.error("ERROR DECODING URI: ", e.message);
+        console.error("ERROR DECODING URI: ", err.message);
     }
 
   return {
--- a/packages/commons-server/src/libs/utils.ts
+++ b/packages/commons-server/src/libs/utils.ts
@@ -112,8 +112,8 @@ export function CreateTransaction(
     try {
         queryString = decodeURI(queryString);
     } catch (err) {
-        queryString = "";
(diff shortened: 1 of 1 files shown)


packages/commons-server/src/libs/utils.ts
```diff
@@ -108,6 +108,13 @@ export function CreateTransaction(
   response: Response
 ): Transaction {
   const requestUrl = new URL(request.originalUrl, 'http://localhost/');
+    let queryString = requestUrl.search.slice(1);
+    try {
+        queryString = decodeURI(queryString);
+    } catch (err) {
+        queryString = '';
+        console.error('ERROR DECODING URI: ', err instanceof Error ? err.message : err);
```
255kb, line 116: One remark: the `console.error` will not log errors using the correct transport (winston for the CLI, and electron-log for the desktop app) and I think info would be lost at least in the desktop app once it's packaged. We could pass the logger instance to the `CreateTransaction` function but it is not always available with the current architecture, especially in the desktop app where the server emit errors rather than logging directly. 
Also, I am not sure logging here bring much value. As a counterexample, when there is a parse error with headers we directly put the error message in the header value, so the user can immediately see the problem in the most obvious place.
Another thing is that when there is a Malformed URI error, the raw query params will be empty, while the structured query params (from Express' `request.query`) will be shown:
image

So, I suggest that we revert to the raw query string in the `catch` maybe prefixed by something like "[Error: URI malformed]":
image
I think there is no reason we should not be able to see the raw version even if malformed. What do you think? 

```suggestion
        queryString = `[Error URI malformed]: ${queryString}`;
```

255kb (changes requested):
Thank you for your contribution! 
I added one remark/change request. 

Also, could you please format using Prettier?

Thank you

255kb:
@tsgodara I will merge this and make the changes as I am going to finalize a new release in the coming days. 
Thank you again for the contribution!

PR #772, opened 2022-07-27
Import and export using CLI

### LINKED ISSUE #678: Ability to export to OpenApi V3 Using the CLI
I cant export my environment to OpenApi V3 (Json) using the CLI

When using the CLI, i'd like the ability to export my environment to OpenApi v3 (JSON), like i can do using the Mockoon UI

Only option currently is using the Mockoon UI (App)

Mavbraz:
Closes #678

**Checklist**
- Import Swagger v2/OpenAPI v3 (JSON or YAML)
  - Required flags: `data` and `output`
  - Optional flags: `port` and `prettify`
  - Notes: If more than one environment is returned by `parseDataFiles` method then a error message will be returned.
- Export environment to OpenAPI v3 (JSON)
  - Required flags: `data` and `output`
  - Optional flags: `type` (`'open-api-v3'` is default)


packages/cli/src/commands/export.ts
```diff
@@ -0,0 +1,54 @@
...
+      required: true
+    }),
+    type: flags.string({
+      char: 't',
+      description: 'Export type',
+      options: ['open-api-v3'],
+      parse: (input) => input.toLowerCase(),
+      default: 'open-api-v3',
+      required: false
+    })
```
255kb, line 31: I am not sure the export `--type` flag is really necessary as OpenAPI v3 is the only format supported. And I honestly don't think we will support v2 as an export option.

packages/cli/src/commands/import.ts
```diff
@@ -0,0 +1,60 @@
+import { Command, flags } from '@oclif/command';
+import { promises as fs } from 'fs';
+import { Messages } from '../constants/messages.constants';
+import { parseDataFiles } from '../libs/data';
+
+export default class Import extends Command {
+  public static description = 'Import a mock API';
```
255kb, line 7: I would be more explicit about the source format:
```suggestion
  public static description = 'Import a OpenAPI v3 specification file (YAML or JSON)';
```

packages/cli/src/commands/import.ts
```diff
@@ -0,0 +1,60 @@
...
+      char: 'o',
+      description:
+        'Generated Mockoon path and name (e.g. `./environment.json`)',
+      required: true
+    }),
+    port: flags.integer({
+      char: 'p',
+      description: 'Environment port',
+      default: 8080
+    }),
```
255kb, line 29: I am not sure having a port flag here is relevant as we can already override the port at runtime.

packages/cli/src/commands/import.ts
```diff
@@ -0,0 +1,60 @@
...
+      char: 'd',
+      description: 'Path or URL to your data file',
+      required: true
+    }),
+    output: flags.string({
+      char: 'o',
+      description:
+        'Generated Mockoon path and name (e.g. `./environment.json`)',
+      required: true
+    }),
```
255kb, line 24: I find these flags a bit confusing as `--data` is for Mockoon's file in all other commands (including the new export). 
Here, it's for the OpenAPI source file. 
I would rather use `--data` as the output and `--input` as the source. 
What do you think?

255kb (changes requested):
Apart from the change requests on the code, here are some general remarks:

- Should we also have a `--prettify` option in th export like you did for the import?
- don't forget to update the documentation in the readme for both commands (you can follow the format of the other commands)

Thank you!

255kb:
Hi @Mavbraz, This PR has no activity since more than one year. Someone else proposed to finish the work (@zerock54). I will hand over this PR next week unless you plan and commit to work on it in the coming weeks. 
Let us know. 
Thanks!

Mavbraz:
Hello, @255kb! Sorry about the delay.
No worries, I'll finish this PR request ASAP.

255kb:
Thanks @Mavbraz , let me know if you need anything.

packages/cli/src/commands/export.ts
Mavbraz: I agree with that point. There's no other sort of export on Mockoon.
I'll remove this flag and if needed it can be added easily without any breaking change.

packages/cli/src/commands/import.ts
Mavbraz: It was just an additional feature that I was thinking about adding, but it does not bring much gain.
I'm going to remove it to keep this command with the same export behavior in the desktop app.

packages/cli/src/commands/import.ts
Mavbraz: I also think this data flag a bit confusing at first.
We can follow the pattern of other applications with input and output flags.
IMO it would be better to keep the output instead of data.
Let me know your thoughts.

packages/cli/src/commands/import.ts
Mavbraz: I will also add that it is possible to import a Swagger v2 file.
What do you think?

Mavbraz:
> Apart from the change requests on the code, here are some general remarks:
> 
>     * Should we also have a `--prettify` option in th export like you did for the import?
> 
>     * don't forget to update the documentation in the readme for both commands (you can follow the format of the other commands)
> 
> 
> Thank you!

I think it's more important having this prettify in the export command.
I'm going to include it as well. (It will have the shortcut "p" since it won't be used for the port flag anymore)

Mavbraz:
It's still missing to update the README and add few tests to this PR.
I'll do both in the next few days.

packages/cli/src/commands/import.ts
255kb: Yes true, import supports both v2 and v3

packages/cli/test/specs/export-command.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+        const jsonFile = await fs.readFile('./tmp/export-legacy-file.json');
+        const jsonFileContent = jsonFile.toString();
+        const expectedFile = await fs.readFile(
+          './test/data/openapi/legacy-exported.json'
+        );
+        const expectedContent = expectedFile.toString();
+
+        expect(jsonFileContent).to.equal(expectedContent);
+      });
+  });
```
255kb, line 160: This test case can be removed after you merge latest main, as the support for legacy data file format has been removed for the next version (see https://github.com/mockoon/mockoon/issues/1138)

packages/cli/test/data/openapi/legacy-exported.json
255kb, line 1: This file can be removed after you merge latest main, as the support for legacy data file format has been removed for the next version (see https://github.com/mockoon/mockoon/issues/1138)

packages/cli/test/specs/export-command.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+        './test/data/legacy-export-file/multi.json',
+        '--output',
+        './tmp/export-multi-legacy-file.json',
+        '--prettify'
+      ])
+      .catch((context) => {
+        expect(context.message).to.equal('Only one environment is allowed');
+      })
+      .it('should fail export');
+  });
```
255kb, line 177: This file can be removed after you merge latest main, as the support for legacy data file format has been removed for the next version (see https://github.com/mockoon/mockoon/issues/1138)

255kb (changes requested):
This looks good. 
I requested some changes as in the meantime I removed support for the legacy file format. You can remove these test cases, it should make the CI pass too. 
Only the readme doc is remaining, and after that we are good to go!

255kb (approved):

255kb:
Thanks for the changes!

PR #782, opened 2022-08-02
Add default path for save and open dialog windows

### LINKED ISSUE #756: Desktop: Change the default dialog folder on a fresh install
On a fresh install of the desktop app, the default dialogs folder is the `~/AppData/Local/Program/Mockoon` folder, at least on Windows. This folder get wiped during each update. So, it's not a safe place to store any data (mock files). 

We need to change the default folder, and maybe add a new settings field to setup the starting folder. 

To reset the dialog previously selected folder (on Windows):
Remove the key in the registry: 
HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\ComDlg32\LastVisitedPidlMRU
(One of the entry is the one from Mockoon)

Hugo-Carbiener:
Closes #756

**Checklist**


255kb:
This is not working on Windows 11 and always opens the `defaultPath` instead of relying on the registry. We may have to create our own settings to save the latest folder where we navigated to.

PR #853, opened 2022-11-02
Add response.statusMessage data information in environment logs

### LINKED ISSUE #852: Add `response.statusMessage` data information in environment logs
### Description:
    As a user of desktop application or CLI version
I want to see status message [status message of response](https://nodejs.org/api/http.html#responsestatusmessage).
Curl verbose output with example, what I'm talking about:
image

### Cause:
Currently I'm using mockoon in Proxy mode for capturing any request to some external service.
That service is using status message as a place of sending information about any error, caused by user, or by server (I mean resposes with 400+ status code). In logs there are no such important information for me, that makes me to retry request manually for getting this status, so my efforts for testing is doubled in that case.

AndreiOrmanji:
Closes #852

I've been added a new property  `statusMessage` in `Transaction.response` located in `packages\commons\src\models\server.model.ts` for getting the statusMessage in "Logs" tab and some other places:
 - `function CreateTransaction` in packages\commons-server\src\libs\utils.ts that converts a pair of `Request` and `Response` objects in a `Transaction` object
 - `EnvironmentLogResponse` in packages\desktop\src\renderer\app\models\environment-logs.model.ts for showing the information from modified `Transaction` in "Logs" tab
 - `function formatLog` in packages\desktop\src\renderer\app\services\data.service.ts for converting the `Transaction` object in 
`EnvironmentLogResponse` object.
- added a new item in "General" section of "Response" tab in packages\desktop\src\renderer\app\components\environment-logs\environment-logs.component.html
- updated existing tests for covering my changes.

**Disclaimer:**
I'm not an experienced open software contributor or at least a JS developer, but I want to help you, Mockoon supporters, to make Mockoon better.

**Checklist**

### DIFF
--- a/packages/desktop/src/renderer/app/components/environment-logs/environment-logs.component.html
+++ b/packages/desktop/src/renderer/app/components/environment-logs/environment-logs.component.html
@@ -328,6 +328,9 @@
                   <div class="environment-logs-content-item">
                     <strong>Status:</strong> {{ selectedLog.response?.status }}
                   </div>
+                  <div class="environment-logs-content-item">
+                    <strong>StatusMessage:</strong> {{ selectedLog.response?.statusMessage }}
+                  </div>
                 </div>
 
                 <!-- Response Headers -->
--- a/packages/desktop/src/renderer/app/components/environment-logs/environment-logs.component.html
+++ b/packages/desktop/src/renderer/app/components/environment-logs/environment-logs.component.html
@@ -326,10 +326,14 @@
                 </div>
                 <div [ngbCollapse]="collapseStates['response.general']">
                   <div class="environment-logs-content-item">
-                    <strong>Status:</strong> {{ selectedLog.response?.status }}
-                  </div>
-                  <div class="environment-logs-content-item">
-                    <strong>StatusMessage:</strong> {{ selectedLog.response?.statusMessage }}
+                    <strong>Status:</strong>
+                    {{
+                      selectedLog.response
+                        ? selectedLog.response.status +
+                          ' - ' +
+                          selectedLog.response.statusMessage
+                        : ''
+                    }}
                   </div>
                 </div>
 
--- a/packages/commons-server/src/libs/utils.ts
+++ b/packages/commons-server/src/libs/utils.ts
@@ -131,6 +131,7 @@ export function CreateTransaction(
     },
     response: {
       statusCode: response.statusCode,
+      statusMessage: response.statusMessage,
       headers: TransformHeaders(response.getHeaders()).sort(AscSort),
       body: DecompressBody(response)
     },
--- a/packages/commons/src/models/server.model.ts
+++ b/packages/commons/src/models/server.model.ts
@@ -43,6 +43,7 @@ export type Transaction = {
   };
   response: {
     statusCode: number;
+    statusMessage: string;
     headers: Header[];
     body: string;
   };
--- a/packages/desktop/src/renderer/app/models/environment-logs.model.ts
+++ b/packages/desktop/src/renderer/app/models/environment-logs.model.ts
@@ -13,6 +13,7 @@ export type EnvironmentLogRequest = {
 
 export type EnvironmentLogResponse = {
   status: number;
+  statusMessage: string;
   headers: Header[];
   body: string;
   truncatedBody?: string;
--- a/packages/desktop/src/renderer/app/services/data.service.ts
+++ b/packages/desktop/src/renderer/app/services/data.service.ts
@@ -81,6 +81,7 @@ export class DataService extends Logger {
       proxied: transaction.proxied,
       response: {
         status: transaction.response.statusCode,
+        statusMessage: transaction.response.statusMessage,
         headers: transaction.response.headers,
         body: transaction.response.body,
         binaryBody: transaction.response.body === BINARY_BODY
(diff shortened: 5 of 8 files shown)


AndreiOrmanji:
I don't know if I'm allowed to mark any item in checklist as "solved", so I've left it all "as is".

packages/desktop/src/renderer/app/components/environment-logs/environment-logs.component.html
```diff
@@ -328,6 +328,9 @@
                   <div class="environment-logs-content-item">
                     <strong>Status:</strong> {{ selectedLog.response?.status }}
                   </div>
+                  <div class="environment-logs-content-item">
+                    <strong>StatusMessage:</strong> {{ selectedLog.response?.statusMessage }}
```
255kb, line 332: Maybe title could be more human readable:
```suggestion
                    <strong>Status message:</strong> {{ selectedLog.response?.statusMessage }}
```

255kb (changes requested):
Thanks for the contribution. 
I run the CI and there are some Prettier formatting issue. Please run Prettier before pushing again. I also add a comment concerning the log label.

255kb:
Now that I am thinking about it, I think it would be better to display the Status message next to the status code, so it's more in line with what other tools like Curl are displaying. What do you think?

AndreiOrmanji:
> Now that I am thinking about it, I think it would be better to display the Status message next to the status code, so it's more in line with what other tools like Curl are displaying. What do you think?

In my optinion,it will "break existing UI consistency" where you have already 'General', 'Headers', 'Body' sections.
If you are ok with adding more details, I can add HTTP version, send in response by server (probably can be useful for other users).

255kb:
> > Now that I am thinking about it, I think it would be better to display the Status message next to the status code, so it's more in line with what other tools like Curl are displaying. What do you think?
> 
> In my optinion,it will "break existing UI consistency" where you have already 'General', 'Headers', 'Body' sections. If you are ok with adding more details, I can add HTTP version, sent in response by server (probably can be useful for other users).

What I meant was to add the message next to the status code to avoid adding a new line, like this:
```
Status: 404 - Not Found
```

Instead of:
```
Status: 404
Status message: Not Found
```
I'm not sure if adding the HTTP version is really useful. Wouldn't it be always something like 1.1 as we are using Node.js' http?

AndreiOrmanji:
> > > Now that I am thinking about it, I think it would be better to display the Status message next to the status code, so it's more in line with what other tools like Curl are displaying. What do you think?
> > 
> > 
> > In my optinion, it will "break existing UI consistency" where you have already 'General', 'Headers', 'Body' sections. If you are ok with adding more details, I can add HTTP version, sent in response by server (probably can be useful for other users).
> 
> What I meant was to add the message next to the status code to avoid adding a new line, like this:
> 
> ```
> Status: 404 - Not Found
> ```
> 
> Instead of:
> 
> ```
> Status: 404
> Status message: Not Found
> ```

Your varians looks good too :)


> I'm not sure if adding the HTTP version is really useful. Wouldn't it be always something like 1.1 as we are using Node.js' http?

It depends on server the requests are sent to. For example, persistent connection (read more at https://developer.mozilla.org/en-US/docs/Web/HTTP/Connection_management_in_HTTP_1.x) is available from 1.1 and higher, so it can be useful to know what version is supported by server.

AndreiOrmanji:
> If you don't mind, maybe display status code and message on one line, I think I prefer 🙂 Also, if you want to display the HTTP version go for it!

I want to add HTTP version. Can you add an example how it would look like?

255kb:
Maybe something like:

```
Protocol: HTTP1.1
Status: 404 - Not found
```

So it's ready for other protocols like web socket **when** we will implement web sockets 😄.

AndreiOrmanji:
> Maybe something like:
> 
> ```
> Protocol: HTTP1.1
> Status: 404 - Not found
> ```
> 
> So it's ready for other protocols like web socket **when** we will implement web sockets smile.

I've implemented like so:
image.

AndreiOrmanji:
Sorry for pipeline fails, but I could not set up project on my machine to execute tests from pipeline...
Fixed problem caused pipeline fail.

255kb:
No worries, I used the CI all the time to run the tests 😅
There are still many test that fail, apparently for an extra ":". 
You can run the test locally by running first `npm run build:desktop:ci` then `npm run test:desktop -- --spec "environment-logs.spec.ts"`.
Let me know if you need help with anything.

AndreiOrmanji:
It seems that node js and express did not provide access to protocol and version to send in response, so I've set values from request, where it was possible.
I would remove `Protocol` section because i'm not sure it was implemented in a "correct" way and to be implemented by someone  who has more experience with JS and node.

I've refused to use docker. With windows I've the following error on test running:
```bash
$ npm -v
8.19.2

$ node -v
v19.0.0

$ nvm -v
Running version 1.1.9.
```
Screenshot:
image

255kb:
> I would remove `Protocol` section because i'm not sure it was implemented in a "correct" way and to be implemented by someone who has more experience with JS and node.

I would say the same. Maybe remove it and stick to the status message only? It's probably too much hassle for the benefits right now.

AndreiOrmanji:
> Maybe remove it and stick to the status message only? It's probably too much hassle for the benefits right now.

Removed changes, related to `Protocol`, so the current version looks like:
image

255kb (approved):

255kb:
Thank you for the changes. Everything looks ok, I will merge and add it to the next release (probably later this month).

AndreiOrmanji:
Thank you very much

Issue #893, opened 2022-12-17
Rule based on data bucket json-path check


dashie:
I have a complex data bucket like this one:

```
{
  "tenants": {
    "smarti": {
      "id": "smarti",
      "sites": {
        "ST0001": {
          "id": "ST0001",
          "name": "Milano",
          "tz": "Europe/Rome"
        },
        "ST0005": {
          "id": "ST0005",
          "name": "Atlanta",
          "tz": "America/New_York"
        }
      },
      "devices": {
        "DV00010591": {
          "id": "DV00010591",
          "name": "Device 1",
          "siteId": "ST0001",
          "feeds": [
            {
              "address": "temp"
            },
            {
              "address": "hum"
            }
          ]
        },
        "DV00010592": {
          "id": "DV00010592",
          "name": "Device 2",
          "siteId": "ST0001",
          "feeds": [
            {
              "address": "temp"
            },
            {
              "address": "hum"
            }
          ]
        },
        "DV00050593": {
          "id": "DV00050593",
          "name": "Device 3",
          "siteId": "ST0005",
          "feeds": [
            {
              "address": "temp"
            },
            {
              "address": "hum"
            }
          ]
        }
      }
    },
    "ioti": {
      "id": "ioti",
      "sites": {
        "ST0025": {
          "id": "ST0025",
          "name": "Firenze",
          "tz": "Europe/Rome"
        },
        "ST0029": {
          "id": "ST0029",
          "name": "Beijing",
          "tz": "Asia/Shanghai"
        }
      },
      "devices": {
        "DV00250534": {
          "id": "DV00250534",
          "name": "Device 4",
          "siteId": "ST0025",
          "feeds": [
            {
              "address": "temp"
            },
            {
              "address": "hum"
            }
          ]
        },
        "DV00250535": {
          "id": "DV00250535",
          "name": "Device 5",
          "siteId": "ST0025",
          "feeds": [
            {
              "address": "temp"
            },
            {
              "address": "hum"
            }
          ]
        },
        "DV00290596": {
          "id": "DV00290596",
          "name": "Device 6",
          "siteId": "ST0029",
          "feeds": [
            {
              "address": "temp"
            },
            {
              "address": "hum"
            }
          ]
        }
      }
    }
  }
}
```

that give me the possibility to mock a consistent set of data.

Now with mockoon it is very easy to create routes like these:

[...]

PR #894, opened 2022-12-17
Add "status" request helper

dashie:

I want to propose this new helper. The impact is very little and it is simple to be used, but can be very useful in many cases.
The idea is to be able to set the response status inside the template.

Imagine a route like this:

    test/:id

and a data bucket with name `testdata` and the following content:

```
{
  "data": {
    "001": {
      "id": "001",
      "name": "Item 1"
    },
    "002": {
      "id": "002",
      "name": "Item 2"
    }
  }
}
```

now using a template like this:

```handlebar
{{setVar 'selection' (concat 'data.' (urlParam 'id'))}}
{{#if (dataRaw 'testdata' @selection true)}}
{{data 'testdata' @selection}}
{{else}}
{{status 404}}
{
  "error": "NOT_EXISTS"
}
{{/if}}
```

we have a routes where we can return 200 with data if the data exists in the bucket

    GET /test/001

or 404 if the data does not exist.

    GET /test/003

To be able to check the path selection I've also added an optional flag on `dataRaw` helper, that return an empty string `""` when the selection path does not match any property.

*If you accept this proposal I can finish it with tests and docs :P*

**Checklist**

- [x] CLI automated tests added (@mockoon/cli)

### DIFF
--- a/packages/commons-server/src/libs/server/server.ts
+++ b/packages/commons-server/src/libs/server/server.ts
@@ -531,6 +531,13 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
         );
       }
 
+      if (request.locals) {
+        const statusCode = request.locals.get('statusCode');
+        if (statusCode) {
+          response.status(statusCode);
+        }
+      }
+
       response.body = content;
 
       response.send(content);
--- a/packages/commons-server/src/libs/server/server.ts
+++ b/packages/commons-server/src/libs/server/server.ts
@@ -633,6 +633,13 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
               request
             );
 
+            if (request.locals) {
+              const statusCode = request.locals.get('statusCode');
+              if (statusCode) {
+                response.status(statusCode);
+              }
+            }
+
             response.body = fileContent;
             response.send(fileContent);
           } catch (error: any) {
--- a/packages/commons-server/src/libs/templating-helpers/data-helpers.ts
+++ b/packages/commons-server/src/libs/templating-helpers/data-helpers.ts
@@ -83,7 +83,11 @@ export const DataHelpers = function (
 
         // ensure a value was found at path
         const foundValue = objectGet(value, path);
-        value = foundValue !== undefined ? foundValue : value;
+        if (parameters.length > 1 && parameters[2]) {
+          value = foundValue !== undefined ? foundValue : '';
+        } else {
+          value = foundValue !== undefined ? foundValue : value;
+        }
 
         return value;
       }
--- a/packages/commons-server/src/libs/templating-helpers/request-helpers.ts
+++ b/packages/commons-server/src/libs/templating-helpers/request-helpers.ts
@@ -27,7 +27,9 @@ export const RequestHelpers = function (
     // set status code
     status: function (status: any) {
       const intCode = parseInt(status, 10);
-      if (isNaN(intCode)) return;
+      if (isNaN(intCode)) {
+        return;
+      }
       if (!request.locals) {
         request.locals = new Map<string, any>();
       }
--- a/packages/commons-server/src/libs/templating-helpers/request-helpers.ts
+++ b/packages/commons-server/src/libs/templating-helpers/request-helpers.ts
@@ -24,6 +24,15 @@ export const RequestHelpers = function (
   environment: Environment
 ) {
   return {
+    // set status code
+    status: function (status: any) {
+      const intCode = parseInt(status, 10);
+      if (isNaN(intCode)) return;
+      if (!request.locals) {
+        request.locals = new Map<string, any>();
+      }
+      request.locals.set('statusCode', intCode);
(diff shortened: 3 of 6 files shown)


255kb:
Thank you for opening this PR. 
I have mixed feelings regarding this feature. As stated in #893, CRUD endpoints will have this kind of behavior. They will be linked to a data bucket and return a 404 when something is missing. The implementation will be similar to what json-server does. 
Introducing this `status` helper is clever, but I have the feeling that it's a bit redundant with the CRUD routes (future) implementation. 
Anyway, the feature is kind of hidden in a helper and it's not adding more complexity to the UI or anything like this. 
So, I guess it can be added 😄.

I will review and probably add some remarks. I saw that the linter is failing. Maybe correct this first so we can run all the tests.

packages/commons-server/src/libs/server/server.ts
```diff
@@ -531,6 +531,13 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
         );
       }
 
+      if (request.locals) {
+        const statusCode = request.locals.get('statusCode');
+        if (statusCode) {
+          response.status(statusCode);
+        }
+      }
+
```
255kb, line 540: This should be replicated in the `sendFile` method too (line 627-628), as the templating system is supported in files.

packages/commons-server/src/libs/templating-helpers/data-helpers.ts
```diff
@@ -83,7 +83,11 @@ export const DataHelpers = function (
 
         // ensure a value was found at path
         const foundValue = objectGet(value, path);
-        value = foundValue !== undefined ? foundValue : value;
+        if (parameters.length > 1 && parameters[2]) {
+          value = foundValue !== undefined ? foundValue : '';
+        } else {
+          value = foundValue !== undefined ? foundValue : value;
+        }
```
255kb, line 90: Could you please put this in a separated PR? I prefer keeping each PR focused on one topic. 
Also, this is clearly a discrepancy between how data/dataRaw and body/bodyRaw helpers are working. So, the change is needed but I would rather think this thoroughly in a separated task. (I can also handle it, in which case you can remove it altogether)

packages/commons-server/src/libs/templating-helpers/request-helpers.ts
```diff
@@ -24,6 +24,17 @@ export const RequestHelpers = function (
...
+    status: function (status: any) {
+      const intCode = parseInt(status, 10);
+      if (isNaN(intCode)) {
+        return;
+      }
+      if (!request.locals) {
+        request.locals = new Map<string, any>();
+      }
+      request.locals.set('statusCode', intCode);
+    },
```
255kb, line 37: As this is the first helper dealing with the response, I would prefer to put this in the helpers.ts file or in a new response-helpers.ts file.

255kb (changes requested):

packages/commons-server/src/libs/server/server.ts
dashie: Ok, no prob, I'm going to add it on sendFile too

packages/commons-server/src/libs/templating-helpers/data-helpers.ts
dashie: So do you want I add this behavior on:
- data/dataRaw
- body/bodyRaw
and that I create a specific new PR for this change?

packages/commons-server/src/libs/templating-helpers/request-helpers.ts
dashie: We need to move also the test in a separate file?

255kb: yes I tried to matche the lib/test filenames

packages/commons-server/src/libs/templating-helpers/data-helpers.ts
255kb: Don't worry, I will do it.
But anyway, PR may be merged but the next release will only occur around January.

dashie: tnx, keep us updated if it is possible

packages/commons-server/src/libs/templating-helpers/request-helpers.ts
dashie: so I wait for your change

255kb: Not sure I understand what you mean. Do you want me to move the code in a new file? Let me know and I will do it.

dashie: Yes, if it is needed and you can do it is great, tnx

255kb:
I realized that this helper is not compatible with the data buckets as they are generated during server start, or during the first call. 
I will finalize this PR, and add some documentation. 
It will be part of the release in a few days.

PR #967, opened 2023-02-27
Feature/962 Use OpenAPI examples

### LINKED ISSUE #962: Use OpenAPI examples
Hello,

Thanks for all the great work on Mockoon, it's an amazing tool.

I'm using Mockoon to generate a mock server for my API, based on an OpenAPI spec file. I have examples for each endpoint response in my spec and I'd like to get one different possible response for each example in Mockoon.

For instance this endpoint has two examples : 

```yaml
  '/categories/{id}/activities':
    get:
      summary: List of activities
      tags:
        - categories
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/activities'
              examples:
                Sport:
                  value:
                    - id: 7
                      name: Basket-ball
                    - id: 8
                      name: Tennis
                Hobbies:
                  value:
                    - id: 9
                      name: Knitting
```

And I'd like to get this result in Mockoon when I import my OpenAPI specification file :  

image

I saw on the discussion #519 that you don't support the `examples` OpenAPI property. Is supporting this property on your roadmap ?

Thanks again and have a great day!

mathilde-lannes:
Closes #962 

When importing an OpenAPI specification file, the `examples` property of the specification is used to generate new route responses. Each example is transformed in a new route response and the default route response is the one generated from the `schema` property.

**Checklist**

### DIFF
--- a/packages/commons-server/src/libs/openapi-converter.ts
+++ b/packages/commons-server/src/libs/openapi-converter.ts
@@ -277,6 +277,10 @@ export class OpenAPIConverter {
                 | OpenAPIV2.SchemaObject
                 | OpenAPIV3.SchemaObject
                 | undefined;
+              let examples:
+                | OpenAPIV2.ExampleObject
+                | OpenAPIV3.ExampleObject
+                | undefined;
 
               if (version === 'SWAGGER') {
                 contentTypeHeaders =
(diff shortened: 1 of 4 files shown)


packages/commons-server/src/libs/openapi-converter.ts
```diff
@@ -297,29 +301,46 @@ export class OpenAPIConverter {
...
-                  routeResponse.headers
-                )
-              });
+              const headers = this.buildResponseHeaders(
+                contentTypeHeaders,
+                routeResponse.headers
+              );
+
+              // add response based on schema
+              if (schema) {
```
255kb, line 318: I think you got this check wrong (the tests in the desktop package broke, initially all the tests where in the desktop app as we don't have any lib). 
The idea was the following: 
Create a mockoon response when there is a status in the OpenAPI spec. Generate a sample body from the schema if any, and take all the other available params: description-> label, status-> statusCode. The response was always created and the schema was subject of a ternary to set the body or not.

As you wrapped everything in an `if`, this response is not generated anymore, and only the default one (line 348) gets generated. But this is supposed to happen only if there is no response at all in the OpenApi spec file.

255kb (changes requested):
Thanks for the contribution and for adding tests in the commons-server package. 
There is a small issue that is breaking the desktop tests.

packages/commons-server/src/libs/openapi-converter.ts
mathilde-lannes: Thanks for pointing that out and for the explanation ! I fixed it.

Issue #996, opened 2023-03-31
Align desktop/CLI/serverless versions 


255kb:
## Problem

Currently, versioning is not optimal and can cause issues.
- Desktop is at v1.x.x and minor is incremented at each release. 
- CLI and serverless packages are incremented following semver depending on the new features/bugfixes added in the shared libs and their own changes (adding a new flag to the CLI, etc). Most of the time, it's a new minor version.

When data need to be migrated, it does not trigger a major change. 
In the desktop app this makes sense as the data are migrated automatically. It's mostly transparent for the user. 
However, CLI and serverless are only compatible with the latest data migration if they integrate the shared library at the correct version. This has two consequences: 
- they need to be synchronized with the installed desktop version when there is a data migration and this can happen on a minor version. Not ideal. 
- synchronizing is not obvious. Minor versions changelogs state something like "CLI v1.2.0 is compatible with desktop v1.25.0" which doesn't make any sense. 
- CLI and serverless packages are often automatically updated as we use most of the time a minor version. 

## Solution

- Align the major version for all the apps/packages: desktop, cli, serverless, commons and commons-server. 
- Follow semver and increment major every time there is a data migration.

**Benefits:** 
- this will also simplify a bit the tagging and release workflow. 
- only one changelog will be written with specific sections for CLI and serverless packages

## TODO

- update the "set version" npm script to set the same version to all the packages
- update the CI workflow libraries trigger tag
- update the contributing guidelines
- update the website to create a new changelog system

Issue #959, opened 2023-02-23
Node.js 14 end of support


255kb:
In april 2023, remove Node.js 14 from supported engines for the CLI/libs.

- [ ] update `engine` from multiple package.json
- [ ] Use Node 16 (or 18, with longer LTS) in all CI workflows
- [ ] update Node.js versions for the CLI test matrix with more recent versions and remove 14
- [ ] update dockerfile node version and dockerize command template

likileads:
Hi @255kb, I wish to work on this issue.
Can you please assign this to me.

255kb:
Sure, thank you :) 
Let me know if you have any question. You can also join our [Discord server](https://discord.com/invite/JmTBk59MTb).

likileads:
Please review my [PR](https://github.com/mockoon/mockoon/pull/998).
Let me know if there's any changes.

255kb:
If you don't mind I will finalize this PR today as it is currently blocking two other. Thanks 🙂

255kb:
This has been released in [v3.0.0](https://mockoon.com/releases/3.0.0/).

Issue #895, opened 2022-12-21
Silent mode


vltansky:
I'd be really happy for silent mode, can be introduec as flag `-s, --silent. I'm using mockoon-cli to run mocks parallel with my client. I'd want to see the output only from my client cli. Probably there are some workarounds to achieve this of course but a silent flag could be a good addition to cli.

255kb:
Could you please elaborate a bit on what you expect from such flag? Would it just suppress all logging sent to the console?

vltansky:
Yes that’s exactly what I would expect. No console output

255kb:
I wonder if this could be solved together with #868 

Using a flag like `--logging file,console`, or `--logging none`. Not sure if it's a standard way to do it. 

Current logging default is: 
- files: always
- console: if `--daemon-off`

vltansky:
yep, sounds good

dosmanak:
I am not sure it is a good idea to disable logging at all.
Especially when you can disable stdout and stderr using `>&/dev/null`. However id there is such need I am not against it.

vltansky:
@dosmanak good point about `>&/dev/null`
changed my package.json script to:

`mockoon-cli start .... &> '/dev/null'`

255kb:
I will close this issue as `> /dev/null` or `> NUL` (Windows) is a common/standard workaround and avoid introducing a new flag. I will update the documentation.

PR #1017, opened 2023-04-20
feat(cli): adding a flag to allow disabling of logs to file

### LINKED ISSUE #868: Add option to disable logging to file
https://github.com/mockoon/mockoon/blob/f765b19164ecfbba86d9cbace598f7024f1383c7/packages/cli/src/commands/start.ts#L171

In case the app logs on stdout, I'd like to disable the logs to file which are in container volume.

martin-bucinskas:

A CLI flag was added to allow disabling of logging to file.
The logs in pm2 are redirected to `/dev/null` when the flag `--disable-log-to-file` or `-X` for short is passed as an argument.

Closes #868

**Checklist**

- [x] CLI automated tests added (@mockoon/cli)


martin-bucinskas:
As far as issue #895 goes, I agree with the comment:
> I am not sure it is a good idea to disable logging at all.
> Especially when you can disable stdout and stderr using >&/dev/null.

Disabling all logging wouldn't be a good idea, and workarounds such as redirecting the output to `/dev/null` when starting mockoon is a viable solution.

packages/cli/src/commands/start.ts
```diff
@@ -167,7 +169,7 @@ export default class Start extends Command {
         : '',
       logTransaction: environmentInfo.logTransaction,
       fileTransportsOptions: [
-        { filename: join(Config.logsPath, `${environmentInfo.name}-out.log`) }
+        { filename: environmentInfo.disableLogToFile ? '/dev/null' : join(Config.logsPath, `${environmentInfo.name}-out.log`) }
       ]
```
255kb, line 173: using `/dev/null` on Windows will actually create a `null` file and still do the logging. 
Keeping the array empty instead should work on all platforms.

packages/cli/src/commands/start.ts
```diff
@@ -196,8 +198,8 @@ export default class Start extends Command {
       min_uptime: 10000,
       kill_timeout: 2000,
       args,
-      error: join(Config.logsPath, `${environmentInfo.name}-error.log`),
-      output: join(Config.logsPath, `${environmentInfo.name}-out.log`),
+      error: environmentInfo.disableLogToFile ? '/dev/null' : join(Config.logsPath, `${environmentInfo.name}-error.log`),
+      output: environmentInfo.disableLogToFile ? '/dev/null' : join(Config.logsPath, `${environmentInfo.name}-out.log`),
```
255kb, line 202: Using '/dev/null' should work with PM2 according to their [documentation ](https://pm2.keymetrics.io/docs/usage/log-management/#disable-logging)

255kb (changes requested):

255kb:
Thank you for the contribution. 
What about making this flag the default in the Dockerfile and when generating one with the `dockerize` command (in docker.constants.ts template)? As discussed on the issue, file logging in Docker doesn't make sense. 

I would even go further and disable file logging when launched as a foreground process. I think there is a way to enable a flag if another one is passed (set `--disable-log-to-file` to true when `--daemon-off` is passed. We wouldn't need to change the Dockerfile template.
Next version will be a major anyway due to multiple changes ([Node.js 14 support drop](https://github.com/mockoon/mockoon/issues/959), and [version alignment between desktop and CLI](https://github.com/mockoon/mockoon/issues/996)). 
What do you think?

packages/cli/src/commands/start.ts
martin-bucinskas: Using `/dev/null` does not create a file on windows on my machine when using PM2.

255kb: yes, it doesn't, it works perfectly. I added this comment in case of doubt, because in the other remark on `fileTransport` using `/dev/null` doesn't work. 
I didn't want you to look into this one also. But I guess my comment was rather confusing... Sorry!

packages/cli/src/commands/start.ts
martin-bucinskas: Unfortunately that is correct, fixed by keeping the array empty when we want to disable logging to a file

packages/cli/src/commands/start.ts
martin-bucinskas: Looks like the interface for this accepts either a string path or undefined, so if we don't want error or output to a file, then we provide undefined, not the prettiest approach, but it seems it will still log to file /dev/null on windows

255kb: According to their to [their documentation](https://pm2.keymetrics.io/docs/usage/log-management/#disable-logging) it either accept `/dev/null` or `NULL` as strings. 
Something like: `error: '/dev/null'`. What you did was correct, as they won't use the path directly.

martin-bucinskas:
> I would even go further and disable file logging when launched as a foreground process. I think > there is a way to enable a flag if another one is passed (set --disable-log-to-file to true when > --daemon-off is passed.

I see where you're heading to, but I don't see a nice way to actually do this. What happens when we actually do want the functionality despite --daemon-off is passed? Do we now need an invert of the log disabling?

I'd rather to just keep both independent of each other and define either or both if needed.

255kb:
> I see where you're heading to, but I don't see a nice way to actually do this. What happens when we actually do want the functionality despite --daemon-off is passed? Do we now need an invert of the log disabling?

True, didn't think of that. Let's keep it like this then.

martin-bucinskas:
@255kb I'll try to resolve the linting issues, but getting inconsistencies on my side vs git actions...
image

255kb:
The linting was ok, but not the formatting. You can use `npm run format:write`.

PR #1060, opened 2023-06-15
Add Netlify instructions to the serverless package readme

roybarber:

Updating the serveless package's instructions with details on how to deploy your mock API to Netlify's serverless

**Checklist**
- [ ] ~~data migration added (@mockoon/commons)~~ (Not Required)
- [ ] ~~data migration automated tests added (@mockoon/commons)~~ (Not Required)
- [ ] ~~CLI automated tests added (@mockoon/cli)~~ (Not Required)
- [ ] ~~desktop automated tests added (@mockoon/desktop)~~ (Not Required)

### DIFF
--- a/packages/serverless/README.md
+++ b/packages/serverless/README.md
@@ -83,6 +83,37 @@ const app = new mockoon.MockoonServerless(mockEnv);
 exports.app = functions.https.onRequest(app);
 ```
 
+### Netlify Functions
+
+To use Mockoon Serverless with Netlify's serverless functions
+
+First create a new Netlify function, if you're not sure how please [read here](https://docs.netlify.com/functions/create/?fn-language=js), with the following code
+
+```javascript
+const mockoon = require('@mockoon/serverless');
+const serverless = require('serverless-http')
+
+// Load the Mockoon Environment object
+const mockEnv = require('./datafile.json');
+
+const mockoonServerless = new mockoon.MockoonServerless(mockEnv)
+
+exports.handler = serverless(mockoonServerless.firebaseApp())
+```
+
+Then you will need to setup a redirect to direct requests to the api. This also means in mockoon you will need to configure the api url setting. In the example we have set the API URL in mockoon the mockoon config to be /api, so in the netlify.toml file add:
+
+```toml
+[[redirects]]
+  force = true
+  from = "/api/*"
+  status = 200
+  to = "/.netlify/functions/NAME_OF_YOUR_FUNCTION/:splat"
+```
+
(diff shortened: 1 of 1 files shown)


255kb (approved):
Thank you for adding this! 
I will add a specific MockoonServerless.netlify() function to avoid any confusion.

255kb:
Ah! this is what I am just reading about. https://docs.netlify.com/integrations/frameworks/express/
It seems to be the recommended way. I will then add a function similar to awsHandler that is wrapping with `serverless-http`.

255kb:
Just one thing, what do you mean by "This also means in mockoon you will need to configure the api url setting" ? 
Does it means that it needs the same API prefix? 
image

roybarber:
Thats right @255kb. We cant use a rewrite like below, since all subsequent pages would attempt to hit the express app.

[[redirects]]
  force = true
  from = "/*"
  status = 200
  to = "/.netlify/functions/mockoon/:splat"
  
So we have to target a specific set of routes by using the prefix. This differs on AWS/Firebase since we get exposed to a completely separate url.

We can do the same on Netlify if the only thing we are deploying is the mock api, then the above config would work, since we are not hosting any other pages or content.  So it doesn't matter if we hit /about as an example, it would just return a 404 from Mockoon

255kb:
Ok, thank you for the clarification. But then, there is no need for an API prefix in Mockoon, or any other setup, as the `splat` will only contain the paths sections following the `/api/`? 
It's not clear for me from what you wrote. I just want to avoid any confusion :)

255kb:
Thanks. 

I tried it myself, and indeed it's working well as long as you have the same prefix in Mockoon, if you remove the prefix it stops working. 
What really confused me, is that the `:splat` is not working as indicated in the documentation. From what I read, it should only append the wildcard (`*`), but it apparently gets ignored with everything being appended after the function name (`/api/anything`), so we need the same prefix in Mockoon. 

After testing, it seems that two things can be done, either add nothing after the function name, and the full path will be appended, or add `:splat` but prefixed, then the splat works. There is a discrepancy between the [documentation](https://docs.netlify.com/integrations/frameworks/express/) (which is using `/.netlify/functions/{function-name}/api/:splat`) and this [blog post](https://www.netlify.com/blog/2021/12/13/setting-up-redirects-on-netlify/) which is using `:splat` only but will still requires a prefix. 

I will update the example to clarify this. Thank you again for your contribution! 

```
[[redirects]]
  force = true
  from = "/api/*"
  status = 200
  to = "/.netlify/functions/{function-name}/:splat"
```

PR #1068, opened 2023-07-05
Adds proxy missing response mode

### LINKED ISSUE #363: Proxy per route based on Rule
We use Mockoon especially to handle the Negative scenarios for the external API Integrations and with the help of mockoon would like handle both positive and negative scenarios based on Rules to forward the request to Proxy or not.

The same way Status, Headers and Responses are provided per Response based on Rules, we can have proxy config as well.

The same kind of feature available in Mock-server and it is very useful.

HarryEMartland:

This PR adds an extra Response Model called PROXY_MISSING. When this mode is enabled on a route if the rules do not match then the route does not action anything and the request continues on to the next route. If proxy mode is enabled the route will then be proxied. 

The solution makes use of expresses next method which is called if no responses are returned from the ResponseRulesInterperter. 

Open for better names for the Response Model as technically if proxy mode is not enabled the request will not be proxied. 

This is a proposed solution for #363

**Checklist**

- [x] commons-server tests added (@mockoon/commons-server)

Closes #363


255kb:
Thank you for contributing. 
The use of the next function is clever, I didn't think about this 🙂

Two remarks: 
1. I am not sure using the `ResponseMode` make sense here as all of its current possible choices disable the rules: 
 
   image
 
   Here, we don't want to disable the rules, but rather forward to the proxy if no rule is fulfiled. I would suggest that we use a response setting instead (wording is just a draft):  

    image
 
   _Note: Using a new property on the `RouteResponse` would require to update the model, schema, default values, and add a migration + migration test in the commons library._

2. In any case, this should be added in the UI somewhere. If we change for no. 1, you can follow how it's done for the `fallbackTo404` property. I know this option would be a bit lost in the settings tab, and I'm open to suggestions. 

I think it's kind of an important feature, so it could be added in a new button with a new icon, after the `ResponseMode` choices:
image

3. Independently if we chose a setting checkbox or a new button, I think there should be some indication that nothing will reach the proxy if one of the `ResponseMode` option is activated. But let's keep this point for later.

255kb:
After discussing, seems that the implementation using the `ResponseMode` is good. 
What remains to be done: 
- Add a fourth option to the ResponseMode selector, with a tooltip
- Choose an icon from [Material icons](https://fonts.google.com/icons): I would suggest ["move down"](https://fonts.google.com/icons?selected=Material+Icons:move_down:&icon.query=move+down), ["Low priority"](https://fonts.google.com/icons?selected=Material+Icons:move_down:&icon.query=low+priority), ["forward"](https://fonts.google.com/icons?selected=Material+Icons:forward:&icon.query=forward) or ["keyboard double arrow right"](https://fonts.google.com/icons?selected=Material+Icons:keyboard_double_arrow_right:&icon.query=keyboard+double+arrow+right)
- add some integration tests, checking if we get a response from a proxied API when no rule is fulfiled. This kind of tests are usually done in the desktop app using webdriver. But they could be in the commons-server package too. 

I have no comment on the current code.

HarryEMartland:
@255kb I have done the remaining work on this. Please could you enable the CI runner as I was struggling to run all the tests on my mac. Thanks

255kb (approved):

PR #1079, opened 2023-07-14
Adds skipIfNoRuleMatch to responses

HarryEMartland:

This PR is an alternative proposed solution for #363 (original pr #1068). This adds an extra setting called `skipIfNoRuleMatch` when this is enabled on the default response and the response is used as a default then the whole route is skipped. This is done using the express next method which triggers express to continue processing routes. If proxy mode is enabled this allows for proxying if rules for a route are not triggered.

**Checklist**

- [x]commons-server tests added (@mockoon/commons-server)

### DIFF
--- a/packages/commons-server/src/libs/response-rules-interpreter.ts
+++ b/packages/commons-server/src/libs/response-rules-interpreter.ts
@@ -40,7 +40,7 @@ export class ResponseRulesInterpreter {
    * Choose the route response depending on the first fulfilled rule.
    * If no rule has been fulfilled get the first route response.
    */
-  public chooseResponse(requestNumber: number): RouteResponse {
+  public chooseResponse(requestNumber: number): RouteResponse | null {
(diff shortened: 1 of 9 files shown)


255kb:
@HarryEMartland Thanks for the changes, Now that I read this I think I misguided in my [previous message](https://github.com/mockoon/mockoon/pull/1068#issuecomment-1632052030). Having this setting at the RouteResponse level can be difficult to setup as the default response can be changed. Then, I think the feature wouldn't work anymore. 
Could you join me on our [Discord server](https://discord.com/invite/JmTBk59MTb) for a quick chat about the implementation? Thanks

255kb:
After discussion, the first implementation seems to make more sense. Closing this in favor of #1068.

255kb:
@ParveeenRao This PR has been closed in favor of #1068.
We usually release a new version every month or two. It may be ready for the september release, but no guarantee.

Issue #988, opened 2023-03-24
Use data buckets in response rules


255kb:
### Discussed in https://github.com/mockoon/mockoon/discussions/986

<div type='discussions-op-text'>

<sup>Originally posted by **Jerantony** March 23, 2023</sup>
It would be great if one could use the data buckets in response rules, so that one can configure for example route parameters which trigger a certain response on the fly. In my case, in a test, I want to receive a 400 response for a route parameter which is randomly generated during the test.

Maybe this could be done either by allowing the data helper in the value of response rules or by including all configured data buckets in comparison operator list of the rule and then using the value as the object path.</div>

Jerantony:
Opened a PR: https://github.com/mockoon/mockoon/pull/1095 🙂

PR #1095, opened 2023-07-26
Allow helpers in response rule value

Jerantony:
**Issue**
https://github.com/mockoon/mockoon/issues/988

**Checklist**

### DIFF
--- a/packages/commons-server/src/libs/response-rules-interpreter.ts
+++ b/packages/commons-server/src/libs/response-rules-interpreter.ts
@@ -163,11 +163,13 @@ export class ResponseRulesInterpreter {
       rule.value = '';
     }
 
+    const parsedRuleValue = this.parseValue(rule.value);
+
     let regex: RegExp;
 
     if (rule.operator.includes('regex')) {
       regex = new RegExp(
-        rule.value,
+        parsedRuleValue,
         rule.operator === 'regex_i' ? 'i' : undefined
       );
 
@@ -177,10 +179,10 @@ export class ResponseRulesInterpreter {
     }
 
     if (Array.isArray(value)) {
-      return value.includes(rule.value);
+      return value.includes(parsedRuleValue);
     }
 
-    return String(value) === String(this.parseValue(rule.value));
+    return String(value) === String(parsedRuleValue);
   };
 
   /**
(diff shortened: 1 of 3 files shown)


packages/commons-server/src/libs/response-rules-interpreter.ts
```diff
@@ -198,4 +203,27 @@ export class ResponseRulesInterpreter {
...
+    try {
+      parsedValue = TemplateParser(
+        false,
+        value,
+        this.environment,
+        this.processedDatabuckets,
+        this.request
+      );
+    } catch (error) {
+      return value;
```
Jerantony, line 224: I thought it makes sense to simply use the unparsed value if parsing fails. Any better idea?

255kb:
Thank you for opening this PR, could you please comment on the #988 issue so I can assign it to you?

packages/commons-server/src/libs/response-rules-interpreter.ts
255kb: No, this makes sense.

packages/commons-server/src/libs/response-rules-interpreter.ts
```diff
@@ -175,7 +180,7 @@ export class ResponseRulesInterpreter {
       return value.includes(rule.value);
     }
 
-    return String(value) === String(rule.value);
+    return String(value) === String(this.parseValue(rule.value));
```
255kb, line 183: I think I would parse the `rule.value` earlier, like on line ~165, so the reex or array include would also check on the parsed value.

255kb (changes requested):
Everything looks good. Thank you for fixing the tests too! 

The only thing I would change, would be to parse the rule.Value a bit earlier so it is used for the other type of rules too.

packages/commons-server/src/libs/response-rules-interpreter.ts
Jerantony: Oh yes, that's a good point!

255kb (approved):

255kb:
It's merged and will be in the next release (probably around september). Thank you!

PR #1173, opened 2023-11-01
response callbacks (mockoon#709)

### LINKED ISSUE #709: Trigger new request using moockoon
Hi @255kb, 

Nowadays is very common async communication between micro services like the scenario bellow:

**AppA** requests **AppB** and **AppB** responds with http 202.
After processing (X seconds), **AppB** responds the process result, on a new request, to **AppA**.

Using moockon to reduce integration test code, I think that could be nicer if it provides some way to trigger or execute that new request to specific host:port and after X seconds, simulating above case.

What do you think?

isuru89:

This PR introduces "callback" mechanism to Mockoon that invokes other external APIs once a mocked route is hit. This is useful to mimic the invocation of webhook behaviour implemented in some APIs. These "callbacks" are different to proxies where the route invokes callback(s) asynchronously and not visible to the callers.
Implementation is very similar to the existing databucket implementation.

***Features***

 * Callbacks can be defined in environment level.
   * Templating is enabled for the calling URL and request body.
 * Callbacks can be duplicated within environment level as well as to other environments.
 * A user can manipulate request body of outgoing callback as a written content, file or as a content from databucket.
 * Callbacks can be attached to routes.
   * Attaching callbacks is optional. 
   * A user can attach zero or more callbacks. There is no limit. 
 * User can delay the invocation by a specified amount of time when attaching a callback to a route.
 * User can navigate to the definition of the attached callback directly from the route/response level.
 * Attached callbacks are invoked in the specified order.
 * Usages of a particular callback can be viewed on the definition level.
   * Clicking on a usage item will redirect to the corresponding router/response level

***Screenshots***

Callbacks can be defined in environment level:
env-callbacks

Callbacks can be attached to routes in response level:
response-callbacks

**Checklist**

- [ ] ~CLI automated tests added (@mockoon/cli)~

Closes #709


255kb:
I started reviewing the feature, please find some remarks below (aside from remarks on the code): 

- I would organize the callback view like the route view, separating the body from the headers in two tabs, for consistency purposes (then you can get rid of the two titles, "headers" and "body"): 

  image
  and
  image

- Use same styling and add a border top to the callback view footer, like in the data bucket view: 
  
  image

- I would also invert the order of the icons in the route's callbacks tab: 
  
  image

Also, nice choice of navigation icon, it's exactly what I had in mind! 

- With very long names, the route's callback list is overflowing: 
  image
I solved this for other dropdown in the latest version. You can see the changes in this [commit](https://github.com/mockoon/mockoon/commit/07bf6efd0a8c49259af2471cbbbfe0902d1eb000). I think the easiest is to use `display: grid` on each line, instead of flex, add a `.mw0` on the `custom-select` and get rid of the `col-x`.

- For the usage tab, I would rather make the responses blue anc clickable, but not the route itself. I find it rather confusing that you can click on all of them, while it make mostly sense to click on the responses, not really on the route. What do you think? 

Maybe something like this? (with `badge badge-custom`):

image




And thank you for making the changes following my previous remarks!

packages/desktop/src/renderer/app/stores/reducer.ts
```diff
@@ -1131,6 +1219,165 @@ export const environmentReducer = (
...
+        'name',
+        'body',
+        'bodyType',
+        'databucketID',
+        'filePath',
+        'headers',
+        'method',
+        'sendFileAsBody',
+        'uri'
+      ];
```
255kb, line 1342: You may have to test a bit, but I think there is no need to mark the environment as "restart needed" as they get directly accessed at the time the call is done. 
The reason why the data buckets mandata a server restart is because they are generated at startup time.

packages/commons-server/src/libs/server/server.ts
```diff
@@ -585,6 +591,100 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+            headers: {}
+          };
+          this.setHeaders(cb.headers || [], sendingHeaders, request);
+
+          setTimeout(() => {
+            fetch(url, {
+              method: cb.method,
+              headers: sendingHeaders.headers,
+              body: content
+            });
```
255kb, line 674: You should deal with fetch Promise with a then and a catch. Otherwise we won't be able to surface errors and it will throw UnhandledPromiseRejectionWarning. 

Something like:

```
fetch(url, {
              method: cb.method,
              headers: sendingHeaders.headers,
              body: content
            })
              .then(() => {})
              .catch((e) => {
                this.emit('error', ServerErrorCodes.CALLBACK_ERROR, e);
              });
```
This helped me surface this error, which is probably why most methods are not working: 
`Request with GET/HEAD method cannot have body.!` for GET or HEAD
Or also: 
`fetch failed!` for a PATCH

Another thing, in the then, a new event could be emitted, like the `transaction-complete`, maybe `callback-executed`, so it would get logged too.

packages/commons-server/src/libs/server/server.ts
```diff
@@ -630,6 +730,116 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+          const buffer = readFileSync(filePath);
+          const form = new FormData();
+          form.append('file', new Blob([buffer]));
+
+          setTimeout(() => {
+            fetch(callback.uri, {
+              method: callback.method,
+              body: form,
+              headers: sendingHeaders.headers
+            });
```
255kb, line 796: Same remarks with the fetch's then and catch.

packages/commons-server/src/libs/server/server.ts
```diff
@@ -630,6 +730,116 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+          if (!definedContentType && fileMimeType) {
+            sendingHeaders.headers['Content-Type'] = fileMimeType;
+          }
+
+          setTimeout(() => {
+            fetch(callback.uri, {
+              method: callback.method,
+              headers: sendingHeaders.headers,
+              body: fileContent
+            });
```
255kb, line 827: Same remarks with the fetch's then and catch.

packages/desktop/src/renderer/app/constants/desktop-messages.constants.ts
```diff
@@ -292,5 +292,10 @@ export const DesktopMessages: {
     logPayload: messageParams,
     toastType: 'success',
     showToast: true
+  }),
+  CALLBACK_ERROR: (messageParams) => ({
+    message: `Error while invoking callback ${messageParams.error}!`,
+    log: true,
+    showToast: false
```
255kb, line 299: You can remove this error message implementation as it will already be logged by the server lib. 
This is only for desktop app specific messages, or when we want to display a toast.

255kb (changes requested):

packages/commons-server/src/libs/server/server.ts
isuru89: Great to know about this!

packages/desktop/src/renderer/app/stores/reducer.ts
isuru89: Yes, you are correct. I removed this after a round of tests.

packages/commons-server/src/libs/server/server.ts
isuru89: I created a new event called `callback-invoked` so that callback invocations will be logged.

isuru89:
With the latest commit, I addressed all the reviews and suggestions you mentioned above.

- ✅ Fixed not being ability to call any HTTP method other than PUT/POST/PATCH. (The reason was that, fetch does not allow specifying `body` field for non-body supporting methods)
- ✅ Reorganized the callback view to have two tabs separately to define the body and headers.
- ✅ Added border top to the ID label
- ✅ Invert the location of go-to callback and delete action inside response callbacks.
- ✅ Fixed the overflow of text for longer callback names
- ✅ Usage tab styling changed to make it differ routes from responses. Also, disabled the ability to click a route.

Thank you @255kb for your guidance.

packages/commons-server/src/libs/server/server.ts
```diff
@@ -630,6 +745,150 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+      } catch (error: any) {
+        fileServingError(error);
+      }
+    } catch (error: any) {
+      this.emit('error', ServerErrorCodes.ROUTE_SERVING_ERROR, error);
+
+      this.sendError(
+        response,
+        format(ServerMessages.ROUTE_SERVING_ERROR, error.message)
+      );
```
255kb, line 888: Same remark, this can be removed.

packages/commons-server/src/libs/server/server.ts
```diff
@@ -630,6 +745,150 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+    invocation: CallbackInvocation,
+    request: Request,
+    response: Response
+  ) {
+    if (!callback.filePath) {
+      return;
+    }
+
+    const fileServingError = (error) => {
+      this.emit('error', ServerErrorCodes.ROUTE_FILE_SERVING_ERROR, error);
```
255kb, line 761: Please create a new error code. Something like CALLBACK_FILE_ERROR, otherwise it will be logged and display a toast in the app like a regular response related file error. 
Like the CALLBACK_ERROR, please include a payload containing the `callbackName` when emitting. 

(same remark for the same error emitted in this function)

255kb (changes requested):

255kb:
Thank you for updating. It's looking great! 
There are some minor things that I didn't notice before, especially regarding the error logging. Let me know what you think. 

## Callback navigation button

Instead of hiding the navigation button when no callback is selected, maybe make it gray, like that there is no UI jump:

This:  
  
image

Becomes this:
  
image

  This would require to add the following scss in the bootstrap-overrides.scss file, inside the `.bt.btn-icon` selector:
  ```scss
  &:disabled,
    &.disabled {
      color: $gray-600;
      pointer-events: none;
    }
  ```

Or an alternative: Always create the entry with the first callback selected by default. It's maybe simpler.

## Unique ID

After thinking about it a bit more, maybe the callback's unique ID (and the callback footer) should be hidden. In the data buckets it make sense to display it as it can be used by the user in a `data` or `dataRaw` helper to reference the bucket. But for the callbacks it's only an internal reference. What do you think?

## Error logging

Aside from the other remarks, a new case in the switch in `listenServerEvents` on line 73 ([link](https://github.com/mockoon/mockoon/pull/1173/files#diff-8d5210da5fc85a292154b9b2a0936a74445fb7aadf420cb0ede236a4d2a77b40L73)) should be added with 'CALLBACK_ERROR' to properly handle the logging message.

When the target server for a callback is down, the message is a bit vague ("fetch failed"). I think all the emitted CALLBACK_ERROR should contain a specific payload containing the `callbackName` at least, so it's easier to find which one failed. You can see a model [here](https://github.com/mockoon/mockoon/pull/1173/files#diff-0795ec89d79ac105c1680f17b98ad74d7bc7ca97fa9d7e6b1916e634accf4f07R909-R912). This `payload` will then be automatically used in the `listenServerEvents` function.

## Full transaction logging option

In another remark I suggested that you only log the full callback if the  `logTransaction` option is enabled. Then, the setting wording in the modal tooltip should be updated a bit: 
image

Maybe: "Logs the full transactions (for requests, responses and callbacks) in the application log files. Greatly increases the volume of logs. Requires a server restart to take effect."

packages/commons-server/src/libs/server/server.ts
```diff
@@ -585,6 +593,113 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+              );
+          }, invocation.latency);
+        }
+      } catch (error: any) {
+        this.emit('error', ServerErrorCodes.CALLBACK_ERROR, error);
+
+        this.sendError(
+          response,
+          format(ServerMessages.CALLBACK_ERROR, error.message)
+        );
```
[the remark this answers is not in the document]
isuru89: You are correct.

isuru89:
With the latest [commit](https://github.com/mockoon/mockoon/pull/1173/commits/9f2a59e8dc830515fb65f590c7def73f842a977c), I addressed the latest review comments made earlier.

✅  Callback navigation button layout fixes. I made automatically set the default value for the very first callback item when a new callback attached to a response.
✅  Also, if there is no callbacks defined when attaching to a response, a message will be shown
✅  Callback id is hidden
✅  Error logging improved as you suggested considering full transaction logging option.

255kb (approved):

255kb:
Thanks for the update @isuru89 
I just finished reviewing. Everything looks good. I will just add two very minor changes after merging. 
It will be included in the next release and I will add some documentation on the website too. 

Thank you again for your contribution.

eliseuborges:
@isuru89 thank you for your amazing work!
@255kb Thank you too to maintain this amazing project.

isuru89:
@255kb 
Great! 
Please, assign me those two minor changes you need. I will happily contribute.  :+1:

255kb:
> @255kb Great! Please, assign me those two minor changes you need. I will happily contribute. 👍

Don't worry, I did it. I realized I forgot to ask you to run the tests migration script, but this script was also broken due to some other modification I did earlier. 
If you want to contribute some more and discuss possible features, feel free to DM me on our [Discord Server](https://discord.gg/FtJjkejKGp).

PR #1212, opened 2023-12-02
Add searching and filtering for CRUD routes with query params

### LINKED ISSUE #1047: Add filtering for CRUD routes with query params
Hello all,
working with routes and **CRUD**, would be nice to be able to apply **filters** to the responses.
Should be posible?
Just a beloved feature request

Thanks for this wonderfull mocker :-)

HighLiuk:
Adds **searching and filtering** functionality to CRUD routes using query params

# Filtering

Adds the ability to use query parameters in the form of `[property]_[operator]`. For instance, if you want to filter an array of objects by their `status` property, you can use:

```
GET /path?status_eq=success
```

With this PR, the following operators are introduced:

| Operator | Description                                   | Example             |
| -------- | --------------------------------------------- | ------------------- |
| `eq`     | Filters by equality                           | `status_eq=success` |
| `ne`     | Filters by inequality                         | `status_ne=error`   |
| `gt`     | Filters by greater                            | `price_gt=10`       |
| `gte`    | Filters by greater or equal                   | `price_gte=10`      |
| `lt`     | Filters by lower                              | `price_lt=10`       |
| `lte`    | Filters by lower or equal                     | `price_lte=10`      |
| `like`   | Filters by partial match \*                   | `name_like=ohn do`  |
| `start`  | Filters properties that start with a value \* | `name_start=john`   |
| `end`    | Filters properties that end with a value \*   | `name_end=doe`      |

\* Case insensitive

- filters work on every primitive value (strings, numbers, booleans, etc.), so you can filter by `true` or `false` values, or even `null` values, e.g. `is_active_eq=true`
- works on arrays of primitives, in which case using `_[operator]` is enough, e.g. `_gte=10`
- supports nesting: `user.email_end=@example.com`
- `like`, `start` and `end` have support for regex out of the box

# Searching

Adds the ability to search an array via the `search` query parameter. Search is a special kind of filter that will look for a partial match in any values (with nesting) in the array. For instance, if you want to search for `john` in an array of users, you can use:

```
GET /path?search=john
```

- works on arrays of primitives (strings, numbers, etc.)

## Meta data

- Adds a new `X-Filtered-Count` header to the response, to indicate the number of items after filtering (not taking into account pagination)

**Checklist**

- [x] Common-Server automated tests added (@mockoon/common-server)
- [x] Desktop automated tests added (@mockoon/desktop)

Closes #1047

### DIFF
--- a/packages/commons-server/src/libs/filters.ts
+++ b/packages/commons-server/src/libs/filters.ts
@@ -5,9 +5,11 @@ const isNum = (value: string) => !isNaN(parseFloat(value));
 
 export const FILTERS = {
   eq: (data, query) =>
-    String(data) === query || (isNum(data) && isNum(query) && data == query),
+    String(data) === query ||
+    (isNum(data) && isNum(query) && Number(data) === Number(query)),
   ne: (data, query) =>
-    String(data) !== query && (!isNum(data) || !isNum(query) || data != query),
+    String(data) !== query &&
+    (!isNum(data) || !isNum(query) || Number(data) !== Number(query)),
   gt: (data, query) => data > query,
   gte: (data, query) => data >= query,
   lt: (data, query) => data < query,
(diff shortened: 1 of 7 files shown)


255kb (approved):
Thank you @HighLiuk nothing much to say :)
Kudos for covering extensively with tests!

PR #1214, opened 2023-12-04
WebSocket support (#83)

### LINKED ISSUE #83: Websocket support
I am currently debugging an token authentication error. The authentication happens via a Websocket connection. Mockoon doesn't allow Websocket request interceptions which would be really dope:)

Ability to also intercept and mock Websocket communication. I think it should be a toggle in the "Environment" settings, just like HTTPS.
A nice bonus would be if you could also define a route to be handled in the Websocket intercepter, but I guess this use case is to specific for my problem.

As an alternative I am mocking the response in the Applications code

**Additional context**
No additional context

isuru89:

After a completion of successful PoC, I thought I would raise this pull request with the initial implementation of the WebSockets.

This implementation covers all basic features of a websocket. Followings are the features and limitations.

**Features**:
 - Can define WebSocket protocol (optionally with TLS) end points and consume them. (e.g. `ws://localhost:3000/<sub-path>`)
 - Can mix both WebSocket and HTTP endpoints together in the same environment.
 - Ability to respond based on the content of incoming message (full-duplex conversations)
 - Ability to broadcast same data to all the connected clients of a particular WebSocket endpoint (broadcast mode)
 - Ability to unicast streaming data for each connected client of a particular WebSocket endpoint (one-to-one mode)

**Limitations**:
  - No Socket.IO support. Only `ws/wss` portocols.
  - Path parameters are not supported for WebSockets
  - No multicast support (this feature is in socket.io with rooms). Only one-to-one and broadcast streaming is supported.
  - No authentications assumed.

**Screenshots**:

Adding a WebSocket route.
mockoon-ws-add

View of the WebSocket route.
mockoon-ws-routes

Shows it in logs.
mockoon-ws-logs

---
**Checklist**

- [ ] ~CLI automated tests added (@mockoon/cli)~

Closes #83


packages/commons-server/src/libs/templating-helpers/web-socket-helpers.ts
255kb, line 1: please reuse the original request helpers' code instead of duplicating

packages/commons-server/src/libs/ws-response-rules-interpreter.ts
255kb, line 1: We should reuse the rules interpreter code too instead of duplicating, otherwise it will open the door to differences in the feature behavior depending on the route type

packages/desktop/src/renderer/app/components/menus/routes-menu/routes-menu.component.html
```diff
@@ -270,6 +273,13 @@
             >
               <app-svg icon="low_priority"></app-svg>
             </div>
+            <div
+              *ngIf="route.type === 'ws'"
+              class="text-primary pe-2 d-flex align-items-center"
+              ngbTooltip="WebSocket Route"
+            >
+              <app-svg icon="websocket"></app-svg>
+            </div>
```
255kb, line 282: I would prefer that we reuse the badge like GET, CRUD, etc with "WS", instead of an icon.
In the style.css $route-badges variables, we could use $indigo that is not yet used.

255kb (changes requested):
Thank you for this huge amount of work!

I started testing and reviewing but encountered some errors that prevented me from continuing further. I will retest as you update the PR.
Aside from the comments I already put on the code, here are some remarks: 

- I think the badge should be reused in the routes menu with a "WS" instead of an icon. Same for the logs menu:
 
image

- in conversation mode, when I was modifying the body, latency, rules, etc. It wasn't applied until I restarted the server. This is not the case with the regular HTTP or CRUD routes. I don't know what is missing in the code, but you will see for the HTTP/CRUD routes a `refreshEnvironment` function being called before each request (https://github.com/mockoon/mockoon/blob/2a0405a12ea5053690cc83ec22d3241d6775fafb/packages/commons-server/src/libs/server/server.ts#L502)

- I'm not sure if the fallback mode make sense for a persisting websocket connection. I think this option should be disabled. 

image

- I think the initial naming of the responses could be kept: "Response 1 {documentation}"

image

- I didn't see the  WS_UNKNOWN_ROUTE happening in the logs when trying to connect on the wrong route

- when activating one of the streaming options (broadcast or one-to-one) I always got the same error message: 

```
{"date":"Tue Jan 16 2024 10:58:16 GMT+0100 (Central European Standard Time)","error":{},"exception":true,"level":"error","message":"uncaughtException: Unexpected end of JSON input\nSyntaxError: Unexpected end of JSON input\n    at JSON.parse (<anonymous>)\n    at parseWebSocketMessage (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\templating-helpers\\web-socket-helpers.js:25:25)\n    at WebSocketHelpers (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\templating-helpers\\web-socket-helpers.js:38:55)\n    at TemplateParser (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\template-parser.js:29:103)\n    at MockoonServer.deriveFinalResponseContentForWebSockets (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\server\\server.js:488:60)\n    at Timeout._onTimeout (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\server\\server.js:511:34)\n    at listOnTimeout (node:internal/timers:569:17)\n    at process.processTimers (node:internal/timers:512:7)","os":{"loadavg":[0,0,0],"uptime":98385.875},"process":{"argv":["c:\\dev\\mockoon\\mockoon\\packages\\desktop\\node_modules\\electron\\dist\\electron.exe",".","--remote-debugging-port=8888","--data-folder=./tmp"],"cwd":"c:\\dev\\mockoon\\mockoon\\packages\\desktop","execPath":"c:\\dev\\mockoon\\mockoon\\packages\\desktop\\node_modules\\electron\\dist\\electron.exe","gid":null,"memoryUsage":{"arrayBuffers":0,"external":3718739,"heapTotal":54034432,"heapUsed":46952552,"rss":155590656},"pid":40784,"uid":null,"version":"v18.16.1"},"stack":"SyntaxError: Unexpected end of JSON input\n    at JSON.parse (<anonymous>)\n    at parseWebSocketMessage (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\templating-helpers\\web-socket-helpers.js:25:25)\n    at WebSocketHelpers (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\templating-helpers\\web-socket-helpers.js:38:55)\n    at TemplateParser (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\template-parser.js:29:103)\n    at MockoonServer.deriveFinalResponseContentForWebSockets (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\server\\server.js:488:60)\n    at Timeout._onTimeout (c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\server\\server.js:511:34)\n    at listOnTimeout (node:internal/timers:569:17)\n    at process.processTimers (node:internal/timers:512:7)","timestamp":"2024-01-16T09:58:16.377Z","trace":[{"column":null,"file":null,"function":"JSON.parse","line":null,"method":"parse","native":false},{"column":25,"file":"c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\templating-helpers\\web-socket-helpers.js","function":"parseWebSocketMessage","line":25,"method":null,"native":false},{"column":55,"file":"c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\templating-helpers\\web-socket-helpers.js","function":"WebSocketHelpers","line":38,"method":null,"native":false},{"column":103,"file":"c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\template-parser.js","function":"TemplateParser","line":29,"method":null,"native":false},{"column":60,"file":"c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\server\\server.js","function":"MockoonServer.deriveFinalResponseContentForWebSockets","line":488,"method":"deriveFinalResponseContentForWebSockets","native":false},{"column":34,"file":"c:\\dev\\mockoon\\mockoon\\packages\\commons-server\\dist\\cjs\\libs\\server\\server.js","function":"Timeout._onTimeout","line":511,"method":"_onTimeout","native":false},{"column":17,"file":"node:internal/timers","function":"listOnTimeout","line":569,"method":null,"native":false},{"column":7,"file":"node:internal/timers","function":"process.processTimers","line":512,"method":"processTimers","native":false}]}
```

Some questions, as I couldn't test everything: 

- in streaming mode, does the rules, random/sequential/disable rules modes work? 
- Shouldn't we put a minimum interval for the streaming mode? I'm afraid if we can setup something really small like 1ms it could crash everything. 
- I think the streaming one-to-one icon could be another one, a bit clearer: maybe [this one](https://fonts.google.com/icons?selected=Material+Icons:multiple_stop:&icon.query=arrow)? We already have it in the app but don't use it anymore. 

Thank you. Let me know on Discord if you want to discuss some of these questions.

isuru89:
To answer your questions.

**1. In streaming mode, does the rules, random/sequential/disable rules modes work?**
     Yes, in streaming mode, one can configure response for each emission randomly, sequentially or always with default response. Rules will be automatically disabled for streaming websockets, because rules will be useful in bidirectional communications, not for unidirectional scenarios like streaming. And I have added a message indicating it.

**2. Shouldn't we put a minimum interval for the streaming mode? I'm afraid if we can setup something really small like 1ms it could crash everything.**
     Great, suggestion. With latest changes addressing all these review comments, I am going to have a lower bound of 10ms for streaming interval, if it is too low. Also, a warning message will be shown to user if user specified a very low value <10ms. 

**3. I think the streaming one-to-one icon could be another one, a bit clearer: maybe [this one](https://fonts.google.com/icons?selected=Material+Icons:multiple_stop:&icon.query=arrow)? We already have it in the app but don't use it anymore**.
    I like the icon you linked in the question and has been changed to it. The app has similar icon under `repeat` but it has  solid arrows. I think dashed/dotted arrows would be perfect.

inuitviking:
Hi!

What's needed to get this completed, @255kb and @isuru89 ?

I'm working for a company that is heavily using the forked version so far (with great success!), and would love to see this come to fruition, and make developing for and with websockets easier. It's already easier as it is now, so we're very thankful for such a contribution to such a wonderful tool! :smile:

I'm not a JS dev, and have never touched electron, but I'm more than willing to test on Windows and Linux, and could provide feedback on that matter. I could look at the code as well, but I fear my lack of experience in JS and its inevitable quirks will be a greater hindrance than a help. :sweat_smile:

255kb:
Hi @inuitviking 
Sorry for not discussing this more publicly (we discussed with @isuru89 on Discord), but after testing this PR, I think there is a major thing missing from the normal ws feature (not the broadcasting mode). When sending messages to the websocket, all the responses are based on the first message, instead of regenerating the template and interpreting the rules each time, like it is right now for the other types of routes, http and crud. I don't see any reason why the bahavior should be different for the ws type. 
Last time I tested (~2 weeks ago) it was still the case.

Again, not talking about broadcasting mode that sends the same message all the time. I think it makes sense. 

@isuru89 I don't think you had time to change this and update your PR? 

As a side note, as this PR is modifying the data model (new route type and options) and because we now have a cloud, and because we don't force updates (on purpose), this must be released in a major version. 
I would like to avoid releasing too frequent major versions like in the past, and release only one per year, maybe two max. This would be the occasion to pack other changes requiring a major version bump.

inuitviking:
Thanks for the swift reply!

> When sending messages to the websocket, all the responses are based on the first message, instead of regenerating the template and interpreting the rules each time, like it is right now for the other types of routes, http and crud. I don't see any reason why the bahavior should be different for the ws type.

That explains the odd behaviour with the websocket when testing, and it working as expected with only the first few messages after a restart of the server. Would be cool with a resolution on that matter! :smile:

> As a side note, as this PR is modifying the data model (new route type and options) and because we now have a cloud, and because we don't force updates (on purpose), this must be released in a major version.
I would like to avoid releasing too frequent major versions like in the past, and release only one per year, maybe two max. This would be the occasion to pack other changes requiring a major version bump.

Completely understandable! Gotta be careful.

I'll be using the fork for now, and if I find the time to look into the code, I'll hop back in here if it's needed. :)

255kb:
> I'll be using the fork for now, and if I find the time to look into the code, I'll hop back in here if it's needed. :)

If you can test more deeply and provide some feedback, that would be really appreciated!

isuru89:
@255kb 
I added that refresh functionality to websocket routes few weeks back. Now the template and route changes should immediately reflect in its responses.

255kb:
> @255kb I added that refresh functionality to websocket routes few weeks back. Now the template and route changes should immediately reflect in its responses.

Thank you. 
There was this refresh thing, but also the fact that each response message should be built based on the entering message, generating a new template and re-interpreting rules like it is right now for the http routes. 
I think for now, the response is always the same after the first message, even if you send a different body. 
Also, I think the rules "request number" should work with ws too (each message could be counted as a request).

packages/desktop/test/specs/ws.spec.ts
```diff
@@ -0,0 +1,238 @@
...
+
+      await ws2.assertReply(
+        '{ "message": "Hello world 3!" }',
+        'ECHO: { "message": "Hello world 3!" }'
+      );
+
+      ws2.close();
+    });
+
+    it('should return correct response for each message', async () => {
```
isuru89, line 81: @255kb 
Isn't this a test case fulfilling what you mean by the other thread? Or, do you have any other different scenario?

255kb: It looks like it. I will retest soon :)

255kb: It's working as you said, great!

255kb (changes requested):
Thanks for updating the PR! It's working quite well. 
I've added some comments, aside from these, here are some things that I think should be changed before merging: 

- The optional response in the Transaction object causes a null pointer (see app logs when switching on the Logs tab) in the environment-logs.component.ts at line 149 as the code expects a response to be present. Aside from that, and to solve this issue, I think the logs' response tab shouldn't be empty and should display at least the status (101) and headers. Later it could display the messages, but I guess it's ok for now to keep the scope of this PR limited. 
- Following a previous discussion on Discord, I understood that broadcasting was not using anything from the request (a `{{body}}` helper would get ignored), but that one to one streaming would use each client's request bodies. But after testing it seems that one to one is not using the bodies. I have the same result than the broadcasting mode. Am I missing something?
- In "normal" ws mode (no streaming/broadcasting), messages are always parsed as strings. I know that this is on purpose, but I think it would be better to try to parse them as JSON. Like that, fetching data in a specific path would work, `{{body 'test'}}` and rules would work too: 
  image

The next release will be a major one (v9.0.0). I target mid-end of August, or end of September. I'm currently dealing with code signing certificates renewal and it could take a bit of time. I would like to add this PR to this update.

packages/commons-server/src/libs/requests.ts
```diff
@@ -0,0 +1,128 @@
...
+    cookies: null,
+    header: (name: string) => req.headers && req.headers[name],
+    get: (headerName: string) => req.headers && req.headers[headerName],
+    hostname: req.headers && req.headers['host'],
+    ip:
+      (req.headers && req.headers['x-forwarded-for']) ||
+      req.socket?.remoteAddress,
+    method: req.method,
+    originalRequest: req,
+    params: {},
```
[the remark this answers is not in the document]
isuru89: By default, there is no way to access cookies directly from a node's inbuilt IncomingMessage object. However, it is possible to parse and store the cookies if specified in header. I will do a simple implementation to parse cookies from headers in next commit.

However, params parsing is a different beast as natively there is no support of parsing path parameters based on patterns like `:paramId`. So, I may have to get help of a third party library, but at the moment, I did not do it. (considering the scope and as well as finding a reliable npm library for it).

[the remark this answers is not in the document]
255kb: I didn't realize requests weren't going through the same process. It makes sense. I would say leave it for now, I'm not sure it's that important.

isuru89:
> * The optional response in the Transaction object causes a null pointer (see app logs when switching on the Logs tab) in the environment-logs.component.ts at line 149 as the code expects a response to be present. Aside from that, and to solve this issue, I think the logs' response tab shouldn't be empty and should display at least the status (101) and headers. Later it could display the messages, but I guess it's ok for now to keep the scope of this PR limited.

This will be fixed

> * Following a previous discussion on Discord, I understood that broadcasting was not using anything from the request (a `{{body}}` helper would get ignored), but that one to one streaming would use each client's request bodies. But after testing it seems that one to one is not using the bodies. I have the same result than the broadcasting mode. Am I missing something?

Current implementation is that streaming begins immediately upon connection establishment, regardless of subsequent incoming messages. Most WebSocket client libraries don't support sending a body with the initial connection request. So, there is no body to work with (for parameters).
Now since you have asked, I wonder should the streaming starts with the connection establishment, or after the very first message received? I believe we should see how users going to use the streaming functionality and extend based on feedback. WDYT?


> * In "normal" ws mode (no streaming/broadcasting), messages are always parsed as strings. I know that this is on purpose, but I think it would be better to try to parse them as JSON. Like that, fetching data in a specific path would work, `{{body 'test'}}` and rules would work too:

Currently it honors content-type header to do the body parsing, but I also think it is better to implement a fallback parsing (as json) when the content-type header is not specified. I will enhance this.

255kb:
> Current implementation is that streaming begins immediately upon connection establishment, regardless of subsequent incoming messages. Most WebSocket client libraries don't support sending a body with the initial connection request. So, there is no body to work with (for parameters). Now since you have asked, I wonder should the streaming starts with the connection establishment, or after the very first message received? I believe we should see how users going to use the streaming functionality and extend based on feedback. WDYT?

I didn't know that, thank you for the explanation. Let's keep it like this and see if there is any specific feedback later. 

> Currently it honors content-type header to do the body parsing, but I also think it is better to implement a fallback parsing (as json) when the content-type header is not specified. I will enhance this.

Now that you mention it, I thought I had the Content-Type set to application/json, but now I see it was missing in my last test. Probably it's a mistake on my side. I think it makes more sense to stick to the Content-Type and not try to parse if there is none. It's currently working like that for the HTTP endpoints. Both features would be aligned.

255kb (approved):

255kb:
Thanks @isuru89 last changes look great 👍 

I will spread the word a bit if people want to test before September release.

nogoodbay:
I have try websocket on local machine, it is so great. but it seems send data to client only after client send data. whether  has a 'send' button which could send data to client after click that button even if that connected client no send any data.

255kb:
@nogoodbay I think you are looking for the broadcasting modes (unicast or multicast): 
image

They should start sending messages as soon as a client connects.

PR #1299, opened 2024-01-30
Overwrite `source` and `pkgrel` in aur package on every package release

Spissable:
Hi, as just discussed via email, here is the PR that addresses the problem with the AUR package update:

The issue was that the initial script had never assumed that the `source` actually changes, aside from the version of course. So once you updated the url with `amd64`, it was only really changed for the checksum generation, but not actually for the `source` field itself. See specifically the line [here](https://github.com/mockoon/mockoon/compare/main...Spissable:mockoon:fix-aur-version-bump?expand=1#diff-38db9d1bd110ea7f464de5a9ef6978ce085254bf68c2cfe1bc3ab6485209ccafR22) which overwrites the `source` field in the `PKGBUILD`.

As a little QoL, I also added the `pkgrel` as a param here. Apparently we were forever stuck with the number `2`, which while not harmful, it wasn't quite right, plus for cases like the one mentioned here, it was nice for me to simply increment it to `3` to indicate that there was an AUR package update addressing an issue, without an actual mockoon update.

Note that the script doesn't need to be run right now, as I already took care of it:
image

### DIFF
--- a/packages/desktop/scripts/aur-version-bump/README.md
+++ b/packages/desktop/scripts/aur-version-bump/README.md
@@ -7,4 +7,6 @@ Set the git configuration to your git email and name in der Dockerfile (line 15,
 If you want to use another ssh key than default you have to pass its value.
 
 # Run:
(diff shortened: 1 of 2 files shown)


255kb (approved):

255kb:
Thank you very much for this fast fix!

PR #1301, opened 2024-02-01
Fix jsonpath return value

### LINKED ISSUE #1297: jsonpath-plus always return an empty array when no matching results are found
**Describe the bug**

When utilizing jsonpath-plus, it has been observed, as reported by @255kb, that the library returns an array containing a single value instead of directly returning the value. This behavior is attributed to the `wrap` option, which is set to `true` by default. 
Read more [here](https://[jsonpath-plus](https://jsonpath-plus.github.io/JSONPath/docs/ts/).github.io/JSONPath/docs/ts/) about it.

This poses a significant challenge, especially when using helper functions with jsonpath, as the expected behaviour is to receive a default value in cases where the path is not found. However, due to the default setting, the functions return an empty array [] instead, preventing the intended default value from being returned.

To tackle this, a breaking change will need to be introduced to set `wrap: false` by default.

**Steps to Reproduce**

Case 1: Single value
Mockoon response
```json
{
  "value": {{body '$.data' '"Default"'}}
}
```

Request body
```json
{
    "data": "value"
}
```

Response body
```json
{
    "value": [
        "value"
    ]
}
```
Expectation
```json
{
    "value": "value"
}
```

Case 2: Invalid json path
```json
{
  "value": {{body '$.invalidData' '"Default"'}}
}
```

Request body
```json
{
    "data": "value"
}
```

Response body
```json
{
    "value": []
}
```
Expectation
```json
{
    "value": "Default"
}
```

**Mockoon version:**
Latest

**OS / OS version:**

Mac OS

ajatkj:

This is a breaking change which affects how jsonPath returns a single value or when it doesn't find any value. Check below issue for more details.

**Checklist**

- [x] lib automated tests added (@mockoon/commons-server)

Closes #1297


255kb:
Thanks for the PR. 
I think the usage of JSONPath plus in the rules interpreter could be updated too. 
When writing the tests for the global variables rule I had a hard time due to the array wrapping thing.

255kb (approved):

PR #1307, opened 2024-02-05
ISSUE-964 Add new two buttons to open file or its directory

### LINKED ISSUE #964: Edit File body type in app
Hello, thank you for this great app!

Could it be possible to edit a file body type directly in Mockoon? (same editor as inline body type would be sufficient)
This would be a great feature as we wouldn't have to find the JSON file we have associated to that response, open that JSON file in any other editor (such as vscode for example) and edit it there. Instead we could simply point to the file in Mockoon, and directly edit it there in-app. It would simplify the editing of a response and improve the experience we have with the app.

This feature might have been already discussed or looked at, but I couldn't find any open issue that had that particular idea or solution.

Edit: Another solution could be to have a "Open file" button next to the "browse" button which would open the file in default editor.

Again, thank you for this great tool, it has been really helpful to us!

Drogonov:

I add new two buttons and add methods in ipc.ts to open files

**Checklist**

Closes #964


Drogonov:
New buttons

Drogonov:
How%20buttons%20works

Drogonov:
How%20buttons%20works

255kb:
Hi @Drogonov 
Thanks for opening this PR. 

I think this is a great feature but I'm not sure adding two more buttons is the best way to go. I would rather add a dropdown menu with the 3 possibilities. 

You could follow how it was done for the routes menu "plus" dropdown:
image
Instead of a square "plus", we could use the [three vertical dots](https://fonts.google.com/icons?selected=Material+Icons:more_vert:&icon.query=dots) (we use Material icons, and this icon is not yet present in the SvgComponent file, we already have a "dots" icon, but they are the horizontal version). 

The menu could includes the 3 entries, with some changes: 

- Current **"Browse"** button (no icon change)
- your new **"Open file in editor"** button (new icon ['file open'](https://fonts.google.com/icons?selected=Material+Icons:file_open:&icon.query=open) icon, as folder with an arrow is currently used for a "move" action)
- your new **"Open folder"** (icon "folder", to be consistent with the environment right click menu entry "Show data file in explorer/finder")

Something like this maybe? 
image
But if we are using the three dots, I think we could get rid of the dropdown caret. 

Last remark, it would be awesome if you could add the same menu to the other file path input in the "Callbacks" view.

Let me know what you think. I'm open to other ideas :)

255kb:
Thank you, 
Do not hesitate to reach me on Discord if you want to discuss something specific. 

Also, you don't have to close the PR and open a new one. You can just push on your branch https://github.com/Drogonov/mockoon/tree/fix/964 and this PR will be updated automatically. It's better, so the discussion stays on the same PR.

Drogonov:
> Thank you, Do not hesitate to reach me on Discord if you want to discuss something specific.
> 
> Also, you don't have to close the PR and open a new one. You can just push on your branch https://github.com/Drogonov/mockoon/tree/fix/964 and this PR will be updated automatically. It's better, so the discussion stays on the same PR.

So I add buttons with icons

- "dots" - for icon with dropdown
- "find_in_page" - for browsing
- "folder_open" - for opening folder
- "folder_move" - for moving to new folder

Pls could you check is it ok?
I didn't find some of the new icons about what you were speaking few messages above

Screenshot 2024-02-27 at 14 38 17

Drogonov:
Screenshot 2024-02-28 at 14 34 37

I have add new icons to design, it guess it looks better now

255kb:
Thanks for the contribution @Drogonov 
As agreed on Discord, I finished the PR with some small modifications (reuse an existing IPC channel, make the paths resolving work on Windows and add the same menu to the callbacks file selector).
It will be part of the March update 🙂

PR #1343, opened 2024-03-17
Data buckets as response rule target

### LINKED ISSUE #1337: Data buckets as response rule target
**Discussed in #1335** 

- The data buckets of the respective folder are made available as targets of response rules
- A new option "Data bucket" is added to the dropdown list of rule targets
- The "property name or path"-field expects an object/JSON path which has to start with the name or id of a data bucket
- The other fields work as usual

Screenshot-Rule-Target-Data-Bucket

Jerantony:

**Checklist**

Closes #1337


255kb (approved):

255kb:
Thanks for the contribution @Jerantony 
I fixed an issue that was broader than this PR. The data buckets containing request helpers (body, etc) are only generated during the first request. 
When such bucket was only used in a rule, it was not generated at all, but that was the case with header values and other fields too.

Jerantony:
Didn't think about this at all, thanks!

PR #1363, opened 2024-04-19
Setting and purging global variables via admin api #1354 

### LINKED ISSUE #1354: Manipulating global env variables via the CLI or admin API
Hi 👋

Great to see the admin API being introduced recently! A suggestion I have is to extend it further by enabling setting global env variables via the API. This would help in cases where one needs to change the mock server behavior at runtime, without restarting the server. Similarly, having an equivalent way to do this via the CLI would be useful too.

Is it something you would consider adding?

sakshi4321:

I have implemented the `setGlobalVars` and `purgeGlobalVars` endpoints for the admin API and introduced two public methods that modify the global variables. Additionally, I updated the ordering of the middleware to accommodate the `parseBody` middleware, which is used to populate ` request.body.`

I have updated the documentation. 

**Checklist**

- [x] CLI tests added (@mockoon/cli)
- [x] desktop UI automated tests added (@mockoon/desktop)

Closes #1354

### DIFF
--- a/packages/commons-server/src/libs/server/admin-endpoint.ts
+++ b/packages/commons-server/src/libs/server/admin-endpoint.ts
@@ -23,37 +23,26 @@ const createStateEndpoints = (
       response: 'State purged successfully.'
     });
   };
-
   const setGlobalEnv = (req, res) => {
-    
     try {
-      
-      const { key, value } = req.body;  // Destructure key and value from req.body
-      
+      const { key, value } = req.body; // Destructure key and value from req.body
       if (key && value) {
-          setGlobalVariables(key, value); // Set global variables, assume this function is defined properly
-
-          res.send({
-              message: `Global variable ${key} has been set to ${value}.`
-          });
+        setGlobalVariables(key, value); // Set global variables, assume this function is defined properly
+        res.send({
+          message: `Global variable ${key} has been set to ${value}.`
+        });
       } else {
-          throw new Error('Key or value missing from request');  // Throw error if key or value are missing
+        throw new Error('Key or value missing from request'); // Throw error if key or value are missing
       }
-  } catch (err) {
+    } catch (err) {
       console.error('Error handling request:', err);
       res.status(400).send({ message: 'Invalid JSON or missing key/value' });
-  }
-   
+    }
   };
-
-
   const purgeGlobalEnv = (req, res) => {
     try {
-      // Assuming 'helpers' is properly imported or defined elsewhere in your code
-      console.log('purging the envs');
-      console.log(purgeGlobalVariables());
       res.send({
-        message: `Global variables have been purged.`
+        message: 'Global variables have been purged.'
       });
     } catch (err) {
       console.log(err);
@@ -61,13 +50,10 @@ const createStateEndpoints = (
       res.status(500).send({ message: 'Failed to retrieve global variable' });
     }
   };
-
-
   app.purge(stateEndpoint, purgeHandler);
   app.post(`${stateEndpoint}/purge`, purgeHandler);
   app.post(`${stateEndpoint}/setGlobalVars`, setGlobalEnv);
   app.post(`${stateEndpoint}/purgeGlobalVars`, purgeGlobalEnv);
-  
 };
 
 /**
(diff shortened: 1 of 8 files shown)


sakshi4321:
@255kb please review this PR. it addresses: #1354

255kb:
Thanks for opening the PR. 

I think I prefered the endpoints paths as you propose in the issue: `/admin-api/global-vars`, so we stick with dash case and have a more RESTy approach. The `/admin-api/state` should stay a route to purge the whole state, not a parent of other resources I think.

I suggest the following paths instead: 

- PURGE /mockoon-admin/state
- POST /mockoon-admin/state/purge
- `POST|PUT|PATCH /mockoon-admin/global-vars`: add/update a variable
- `PURGE /mockoon-admin/global-vars`: remove all the global vars
- `POST/mockoon-admin/global-vars/purge`: remove all the global vars --> the reason for this second endpoint is that PURGE method is often filtered by web application firewalls. 

Could you also add the purging of the global variables to the main `PURGE /admin-api/state` route? 
I will add more remarks on the code itself. 

Thank you!

packages/commons-server/src/libs/server/admin-endpoint.ts
```diff
@@ -1,43 +1,88 @@
...
+      if (key && value) {
+        setGlobalVariables(key, value); // Set global variables
+        res.send({
+          message: `Global variable ${key} has been set to ${value}.`
+        });
+      } else {
+        throw new Error('Key or value missing from request'); // Throw error if key or value are missing
+      }
+    } catch (err) {
+      console.error('Error handling request:', err);
```
255kb, line 41: the console.error will be lost, you should use the logger instead. But I think the call response is enough and we could remove this.

packages/commons-server/src/libs/server/admin-endpoint.ts
```diff
@@ -1,43 +1,88 @@
...
+  };
+  const purgeGlobalVars = (req, res) => {
+    try {
+      purgeGlobalVariables();
+      res.send({
+        message: 'Global variables have been purged.'
+      });
+    } catch (err) {
+      console.log(err);
+      res.status(500).send({ message: 'Failed to retrieve global variable' });
```
255kb, line 53:
```suggestion
      res.status(500).send({ message: 'Failed to purge global variables' });
```

packages/commons-server/src/libs/server/server.ts
```diff
@@ -197,6 +197,22 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
+    this.globalVariables[key] = value;
+  };
+
+  /**
+   * Resets the global variables
+   */
+  public purgeGlobalVariables = (): void => {
+    Object.keys(this.globalVariables).forEach(
+      (key) => delete this.globalVariables[key]
+    );
```
255kb, line 213: This could be simplified to:
```suggestion
    this.globalVariables= {};
```

packages/commons-server/src/libs/server/server.ts
```diff
@@ -214,26 +230,31 @@ export class MockoonServer extends (EventEmitter as new () => TypedEmitter<Serve
...
 
-        // reset databuckets
-        this.processedDatabuckets = [];
-        this.generateDatabuckets(this.environment);
-      }
-    });
+          // reset databuckets
+          this.processedDatabuckets = [];
+          this.generateDatabuckets(this.environment);
+        }
```
255kb, line 248: Could you add a call to `this.purgeGlobalVariables` here, so the full state is purged?

packages/commons-server/src/libs/server/admin-endpoint.ts
```diff
@@ -1,43 +1,88 @@
...
+    }
+  };
+  const purgeGlobalVars = (req, res) => {
+    try {
+      purgeGlobalVariables();
+      res.send({
+        message: 'Global variables have been purged.'
+      });
+    } catch (err) {
+      console.log(err);
```
255kb, line 52: the console.error will be lost, you should use the logger instead. But I think the call response is enough and we could remove this.

255kb (changes requested):

255kb:
Just saw you updated the PR, thanks. But why remove the PUT and PATCH routes? They were good I think! Even if they are technically doing the same thing as the POST route, it's better to follow REST semantics.

255kb:
Don't bother about macos tests not passing, they are flaky since the latest Electron/webdriverio update. The two other (win/linux) are passing so it's good.

sakshi4321:
@255kb I have updated the docs and the API endpoints

sakshi4321 (commented):
Added admin api endpoints , tests

255kb:
Perfect, thanks for the changes. This will be in next week release.

PR #1366, opened 2024-04-22
Issue #1365 - add warning when json is invalid

### LINKED ISSUE #1365: Json validation on beautify
Currently there is no json validation when beautifying json responses - only a comment that beatification should not occur in the catch

  public formatBody() {
    const activeRouteResponse = this.store.getActiveRouteResponse();

    if (!activeRouteResponse.body) {
      return;
    }

    const contentType = GetRouteResponseContentType(
      this.store.getActiveEnvironment(),
      activeRouteResponse
    );

    if (stringIncludesArrayItems(ParsedJSONBodyMimeTypes, contentType)) {
      try {
        this.activeRouteResponseForm
          .get('body')
          .setValue(
            JSON.stringify(
              JSON.parse(activeRouteResponse.body),
              undefined,
              INDENT_SIZE
            )
          );
      } catch (e) {
        **// ignore any errors with parsing / stringifying the JSON**
      }
    }
  }

avollmaier:

Currently there is no json validation when beautifying json responses - only a comment that beatification should not occur in the catch

Closes #1365

### DIFF
--- a/packages/desktop/src/renderer/app/components/environment-routes/environment-routes.component.ts
+++ b/packages/desktop/src/renderer/app/components/environment-routes/environment-routes.component.ts
@@ -496,7 +496,13 @@ export class EnvironmentRoutesComponent implements OnInit, OnDestroy {
             )
           );
       } catch (e) {
-        // ignore any errors with parsing / stringifying the JSON
+        this.uiService
+          .showConfirmDialog({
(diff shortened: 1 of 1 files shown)


255kb:
See my comment on the [parent issue](https://github.com/mockoon/mockoon/issues/1365)

PR #1436, opened 2024-06-29
Add JS templating in responses. Closes 1435, 445.

### LINKED ISSUE #445: Option to define custom handlebars helpers
To simplify the template JSON or to enhance the ability to process the data(like header or query)

Maybe we could add a entry to write some JS code to define the helper we want.
In this JS file, we can access the base helper function, like `faker`, and the base data, like `header` of the request.
Then we could write some functions to do what we want.
After that, we export the functions or we registered the helpers manually.
Finally, we could use them in the template.

### LINKED ISSUE #1435: Support Javascript as a templating language
In some cases producing dynamic responses with Handlebars is cumbersome. This can be especially true when creating JSON responses, where you are manually managing things like commas.

Javascript is better suited to this task. The full power of JS scripting is available to assemble an object, and then it can simply be stringified to produce the response.

Existing Handlebars helpers can be bridged so that they can be directly invoked from javascript, but maintained as a single collection as they are now, providing the full capability of Mockoon's templating to JS in addition to Handlebars.

The Handlebars templating library can also be exposed to Javascript, enabling custom helpers to be added, addressing #445 .

digicide:

- Detects user intent to utilize Javascript with a magic comment in the response: `//JS` or `// javascript`.
- Uses a `node:vm` with an isolated sandbox to execute the Javascript
- Injects Mockoon's built-in handlebars helpers, each wrapped with an interface to allow calling as a Javascript function from within the script. Ordered and named parameters are bridged, and SafeStrings are unwrapped in the return value. This collection of helpers is exposed as `mockoon` in the global scope within the script. This should be self maintaining: as new helpers are added they will be automatically available to both Handlebars and Javascript.
- A siloed Handlebars instance is also exposed to the script in the `handlebars` global variable. Mockoon's unmodified Handlebars helpers are exposed as `helpers`. Users can use Javascript to compile template strings and pass in Mockoon's built-in helpers, along with custom helpers defined in the script. This addresses #445 .
- A global variable `result` is available to the script. The user must assign the final value for the response to this variable, from which the template parser consumes it.
- Results of type `string` are used as-is. Object types are stringified, allowing for powerful generation of JSON responses.

Example of invoking Javascript in a response:
```javascript
// javascript
result = Array.from(Array(10).keys());
```

Example of invoking Mockoon functionality in Javascript:
```javascript
// javascript
result = mockoon.queryParam('str');
```

Example of invoking Mockoon functionality with named parameters in Javascript:
```javascript
// javascript
result = mockoon.dateTimeShift({
  date: "2021-01-01",
  format: "yyyy-MM-dd HH:mm:ss",
  years: 1,
  months: 1,
  days: 1,
  hours: 1,
  minutes: 1,
  seconds: 1,
});
```

Example of using Javascript to invoke Handlebars:
```javascript
// javascript
const templateStr = '{{queryParam "this is a test"}}';
const template = handlebars.compile(templateStr);
result = template({}, { helpers });
```

Example of using Javascript to invoke Handlebars while adding a custom helper:
```javascript
// javascript
helpers.camelcase = (str) => str.toString().toLowerCase().replace(
  /[^a-zA-Z0-9]+(.)/g,
  (m, chr) => chr.toUpperCase()
);
const templateStr = '{{camelcase (queryParam "this is a test")}}';
const template = handlebars.compile(templateStr);
result = template({}, { helpers });
```

**Checklist**

Closes #1435 
Closes #445

### DIFF
--- a/packages/commons-server/src/libs/template-parser.ts
+++ b/packages/commons-server/src/libs/template-parser.ts
@@ -1,6 +1,11 @@
 import { Environment, ProcessedDatabucket } from '@mockoon/commons';
 import { Request, Response } from 'express';
-import { compile as hbsCompile } from 'handlebars';
+import {
+  SafeString,
+  compile as hbsCompile,
+  create as hbsCreate
+} from 'handlebars';
+import vm from 'node:vm';
 import { DataHelpers } from './templating-helpers/data-helpers';
 import { FakerWrapper } from './templating-helpers/faker-wrapper';
 import { GlobalHelpers } from './templating-helpers/global-helpers';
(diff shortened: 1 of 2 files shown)


digicide:
Per contribution guidelines, asking a maintainer to review the code. @ajatkj can you review?

digicide:
Unfortunately I think you are correct. `node:vm` offers no protection against untrusted javascript. I don't the the issue isolated-vm has with electron would be an issue for this project, but it (and a couple other solutions) provide asynchronous evaluation, so would require a bigger refactor to integrate.

The best option will be the ShadowRealm proposal which is in stage 3, but I'm unable to determine when it might be coming to Node. Until then, the options might be to drop the feature, or present users with a warning/confirmation when they run JS (similar to how VS Code asks whether you trust the authors of a project you open).

255kb:
> @255kb Unfortunately I think you are correct. `node:vm` offers no protection against untrusted javascript. I don't the the issue isolated-vm has with electron would be an issue for this project, but it (and a couple other solutions) provide asynchronous evaluation, so would require a bigger refactor to integrate.
> 
> The best option will be the ShadowRealm proposal which is in stage 3, but I'm unable to determine when it might be coming to Node. Until then, the options might be to drop the feature, or present users with a warning/confirmation when they run JS (similar to how VS Code asks whether you trust the authors of a project you open). Can you advise on what would be acceptable?

It's a good question and I'm not sure I have a definitive answer to this. 
Now that we have a cloud and people are increasingly sharing their mocks, the security risk is real. We recently had a similar issue due to JSONPath Plus using `eval`. And after working on it with the security researcher, it appears that mitigating RCE vulnerabilities is not trivial. 
Another thing I would like to avoid is to have features unavailable in some environments (cloud) or under some conditions (like requiring the user to trust before running). Since the beginning, all the features are working in the same way in the desktop app, CLI, etc. and the cloud too (even if it takes sometimes a bit more time to make it available there), without having to think about it. 

I would be tempted to drop the feature. On the other hand, it would be really nice to have this, but I think it would require more research/testing. Maybe using a parser that only support a subset of the JS syntax ([like JSONPath did recently, using jsep](https://github.com/JSONPath-Plus/JSONPath/blob/main/CHANGES.md)). But in this case the question would be what t o allow and where to stop...

digicide:
That's fair, please drop the feature. I will reopen if better options become available.

255kb:
Thank you for yoru work @digicide 
As you said, let's revisit when there is a safer option.

PR #1448, opened 2024-07-16
Add array include rule

### LINKED ISSUE #1444: Add an "array includes" rule
## Problem

It's currently not possible to check in rules that an array (in a bucket, or in the body) includes a specific value. 

## Solution

Add a new "Array includes" rule that will check if an array contains a specific value. It could revert to equals if the provided target is not an array.

ajatkj:

Per bug #1444 added new rule - "array includes"

**Checklist**

Closes #1444


255kb (changes requested):
Thanks for opening the PR @ajatkj 

Some remarks: 

- the environment-schema.constants.ts mus tbe updated to account for the new operator: 
```typescript
  operator: Joi.string()
    .valid('equals', 'regex', 'regex_i', 'null', 'empty_array', 'array_includes') <-- here
    .failover(ResponseRuleDefault.operator)
    .required()
```
- the `array_includes` rule should be disabled like `null` and `empty_array` when "Route parameter" or "Request number" is selected
image

packages/commons-server/src/libs/response-rules-interpreter.ts
```diff
@@ -170,6 +170,10 @@ export class ResponseRulesInterpreter {
 
     const parsedRuleValue = this.parseValue(rule.value);
 
+    if (rule.operator === 'array_includes' && rule.modifier) {
+      return Array.isArray(value) && value.includes(parsedRuleValue);
```
255kb, line 174: This currently works only with arrays of strings as the rule's value field is always containing a string. 
Could you make it work with numbers, booleans and null too?

If could be something like this (not tested): 
`return Array.isArray(value) && value.some(element => String(element) === parsedRuleValue);`

255kb (changes requested):

packages/commons-server/src/libs/response-rules-interpreter.ts
ajatkj: Noted

255kb (approved):

255kb:
Thanks @ajatkj for the changes!

PR #1555, opened 2024-10-14
Feature/1092 json schema validator

### LINKED ISSUE #1452: Request payload validation
### Discussed in https://github.com/mockoon/mockoon/discussions/1434

<div type='discussions-op-text'>

<sup>Originally posted by **sebastien-savalle** June 29, 2024</sup>
Hi,

Mockoon is a very useful tool for prototyping and to run tests when we don't have access to remote system.
Sometimes, we have integration issues because we are validating requests have been correctly sent to mockoon and the response is valid. But, we are validating the request payload sent to mockoon.

I would like to ensure that the request body is valid, and optionally return a bad request when this is not compliant with a given schema.
I would be nice to integrate AJV for example. https://ajv.js.org/

What do you think ?
Regards,
Sebastien</div>

ajatkj:
Implement json schema validation in response rules using 'ajv' library using new modified - `valid json schema`.
The schema is provided via databuckets according to JSONSchema standard.
There is an option to select only a part of schema using object-path to retrieve the subschema.

**Checklist**

Closes #1452


255kb (approved):
Thanks @ajatkj for the PR, everything looks good!

PR #1604, opened 2024-11-22
support tls client auth

### LINKED ISSUE #1600: Can we support client Authentication as well?
For testing complex mutual TLS mocks, can we also support requestCert option?

This can be enabled as part of the tlsOption parameter - requestCert - true or false

krnbr:

This will make changes to the UI and server to support the **mutual TLS** aka **client auth** aka **MTLS** aka **MASSL**

Like below:-

image

**Checklist**

- [x] CLI tests added (@mockoon/cli)

Closes #1600

### DIFF
--- a/packages/cli/test/data/envs/file.json
+++ b/packages/cli/test/data/envs/file.json
@@ -53,7 +53,8 @@
     "certPath": "",
     "keyPath": "",
     "caPath": "",
-    "passphrase": ""
+    "passphrase": "",
+    "requestCert": false
   },
   "data": [],
   "folders": [],
--- a/packages/cli/test/data/envs/mock-export.json
+++ b/packages/cli/test/data/envs/mock-export.json
@@ -51,7 +51,8 @@
     "certPath": "",
     "keyPath": "",
     "caPath": "",
-    "passphrase": ""
+    "passphrase": "",
+    "requestCert": false
   },
   "data": [],
   "folders": [],
--- a/packages/cli/test/data/envs/mock1.json
+++ b/packages/cli/test/data/envs/mock1.json
@@ -84,7 +84,8 @@
     "certPath": "",
     "keyPath": "",
     "caPath": "",
-    "passphrase": ""
+    "passphrase": "",
+    "requestCert": false
   },
   "data": [],
   "folders": [],
--- a/packages/cli/test/data/envs/mock1noext
+++ b/packages/cli/test/data/envs/mock1noext
@@ -53,7 +53,8 @@
     "certPath": "",
     "keyPath": "",
     "caPath": "",
-    "passphrase": ""
+    "passphrase": "",
+    "requestCert": false
   },
   "data": [],
   "folders": [],
--- a/packages/cli/test/data/envs/mock2.json
+++ b/packages/cli/test/data/envs/mock2.json
@@ -53,7 +53,8 @@
     "certPath": "",
     "keyPath": "",
     "caPath": "",
-    "passphrase": ""
+    "passphrase": "",
+    "requestCert": false
   },
   "data": [],
   "folders": [],
--- a/packages/cli/test/data/envs/mock3.json
+++ b/packages/cli/test/data/envs/mock3.json
@@ -53,7 +53,8 @@
     "certPath": "",
     "keyPath": "",
     "caPath": "",
-    "passphrase": ""
+    "passphrase": "",
+    "requestCert": false
   },
   "data": [],
   "folders": [],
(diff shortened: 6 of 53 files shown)


255kb:
Thanks for opening an issue+PR. 
One remark: I'm under the impression that mutual TLS is not the most used API authentication out there (probably the least). 
I'm not saying there is no value in what you worked on, but I don't think it is a priority. 

Also, as much as I value security, I don't really see the purpose of enabling such a strong authentication method in the desktop application that is mostly used on localhost. Especially if `rejectUnauthorized` is enabled, which make the authentication completely optional.

For these reasons, and because such change would require a major release, I don't think we would ever add this in the desktop app. 

I could see the value of such feature as a runtime option in the CLI for example, which is deployed somewhere authentication could be needed. It could take the form of a flag `--request-cert` which would enable the option.

krnbr:
we at my organization are protecting the APIs using MASSL/ MTLS heavily.

we used mockoon heavily for the development on dev machines, to validate the MTLS based flow, this was a missing feature for sure.

I hope it can be of great importance for enterprise users for sure. I will take a look about `rejectUnauthorized` flag, in detail.

krnbr:
> Thanks for opening an issue+PR. One remark: I'm under the impression that mutual TLS is not the most used API authentication out there (probably the least). I'm not saying there is no value in what you worked on, but I don't think it is a priority.
> 
> Also, as much as I value security, I don't really see the purpose of enabling such a strong authentication method in the desktop application that is mostly used on localhost. Especially if `rejectUnauthorized` is enabled, which make the authentication completely optional.
> 
> For these reasons, and because such change would require a major release, I don't think we would ever add this in the desktop app.
> 
> I could see the value of such feature as a runtime option in the CLI for example, which is deployed somewhere authentication could be needed. It could take the form of a flag `--request-cert` which would enable the option. It would also be a much lighter change.

Thanks for replying :-) 

we at my organization are protecting the APIs using MASSL/ MTLS heavily. Upon a click of a flag making tls disabled or enabled makes a huge difference for sure!

we used mockoon heavily for the development on dev machines, to validate the MTLS based flow, this was a missing feature for sure.

I feel, it can be of great importance for enterprise users for sure. I will take a look about `rejectUnauthorized` flag, in detail.

I showcased among a closed group of devs, they were impressed!

ScottChapman:
I am in a similar (but possibly different) situation where our (pretty large global company) has a best practice of protecting all our APIs with client certs. As stated above it isn't so much an issue for running the mocks per se.

But for me it is a blocker for being able to setup a proxy to an existing API as there is no way to specify client certs to do that.

255kb:
Thanks @ScottChapman for the feedback. This PR is not dealing with this topic for the proxy. I think it should be handled ina separated discussion evne if I suspect solving one could help solve the other (I see you opened another [discussion](https://github.com/mockoon/mockoon/discussions/1609)).

@krnbr I'm not denying the fact that it could be useful for you and your company. I'm just saying that I don't think it's a priority, I don't see the point of enabling this for localhost dev (or I'm missing something), it requires a change in the UI + a major release which is a lot for a such a small feature. We cannot add all the features or it will soon become the next Jira, both loved and dreaded. 

But now, I got an idea. Couldn't we enable `requestCert` all the time together with `rejectUnauthorized: false`. Like this, you would be able to simulate mutual TLS without it being mandatory. 
It would also not require a major update nor a UI/data model change. 
I just don't know if there is any unwanted side effect.

krnbr:
@255kb do you mind if I create a blog relating to this PR and observe people's feedback?

255kb:
> @255kb do you mind if I create a blog relating to this PR and observe people's feedback?

Sure!

tknerr:
Just to chime in here: I'm also working for a client where we want to mock an API and also test the mTLS connection, stumbled across Mockoon today (and looks really nice!), and finally ended here when checking for mTLS support.

The other candidate we are currently looking at is MockServer, [which supports mTLS](https://www.mock-server.com/mock_server/HTTPS_TLS.html), but seems like no longer actively maintained for the past 2-3 years, so honestly Mockoon looks a bit more appealing and providing better developer experience from what I have seen so far ;-)

To me personally, having mTLS support in the CLI is more important, as it would allow us to spin up a mock API via mockoon/cli:latest docker image or similar on our CI agents. Having the same feature exposed via the UI makes sense to me as well, but for the use case in the CI it's always CLI first...

255kb:
Closing this as in its current form it's not the best implementation of mTLS. If we support mTLS in the future it will probably be on a per-deployment (Docker/CLI) rather than through the data model.

PR #1988, opened 2025-12-02
show variants in route select for easy selection

sdg9:

**Checklist**

image

Closes #{issue_number}


255kb:
@sdg9 Thank you for opening a PR. 

While I understand the idea of having a way to temporarily override some settings at runtime without persisting them, I don't think managing that from the UI is the right approach. I can only see confusion arising from this and it's rather the role of the [admin API](https://mockoon.com/docs/latest/admin-api/environment-configuration-update/) exposed by each running server, even if there are not a lot of granular endpoints to manage the mock yet. Finally, it adds more places where the same option can be triggered/modified which is not ideal in term of UX. 

_Note: We usually don't accept PRs that were not discussed in an issue beforehand (see contributing.md for more details)._

PR #2091, opened 2026-02-08
Add recently used commands to command palette

### LINKED ISSUE #2036: Add recently used commands to the command palette
## Problem to solve

Command palette has no list for recently used commands.

## Proposed solution

Add a list of recently used commands at the top of the list followed by a separator.
- Limit to the last 10. 
- Save it in the settings.

rohanakode490:

This PR implements a "Recently Used" commands feature in the Command Palette. It allows users to quickly access their most frequent actions, which remain pinned to the top even during searches (following the VSCode approach).

**Key Changes:**
* **Persistent Storage:** Added `commandPaletteRecentCommands: string[]` to the global `Settings` model and validation schema to persist the last 10 unique commands.
* **Usage Tracking:** Implemented a `recordUsage` method in `CommandPaletteService` triggered when a command is activated via the palette.
* **Architectural Refactor:** Separated command generation (generateCommands) from history mapping (getRecentCommands) to keep the service logic thin and predictable.
* **Dual-Section Filtering (VSCode-style)**:
  -   The palette now displays two sections: **Recent** and **Other Commands**.
  -   Both sections are filtered simultaneously during search.
  -   Matching recent commands stay pinned at the very top of the results.

* **UI Enhancements:** Added smart headers and separators that only appear when their respective sections contain matching items.
* **Technical Fixes:** Updated Angular track logic and HTML id attributes to ensure unique DOM elements when a command appears in both sections.

**How to verify:**
1. Open Command Palette (**Ctrl+P** or **Cmd+P**).
2. Search for and execute a command (e.g., "Zoom In").
3. Re-open the palette; the command should now appear in a new "Recent" section at the top.
4. Type a search query; the recent command should stay pinned at the top if it matches the query.
5. Verify the history survives an application restart.

**Documentation:**
- Limit: Last 10 unique commands.
- UI: Included "Recent" and "Other Commands" headers.

**Video:**
The following is a video which shows:
1. **Usage Recording:** Executing a command (e.g., "Zoom In") immediately adds it to the "Recent" section at the top of the palette.
2. **Smart Filtering:** When searching, the "Recent" header automatically disappears if no recently used commands match the search query.
3. **History Management:** Once the history reaches 10 commands, the oldest command is removed to keep the user settings clean and the list focused.

https://www.loom.com/share/cca2806ff56b44de889119a0248f223f

**Checklist**

Closes #2036

### DIFF
--- a/packages/app/src/renderer/app/components/modals/command-palette-modal/command-palette-modal.component.html
+++ b/packages/app/src/renderer/app/components/modals/command-palette-modal/command-palette-modal.component.html
@@ -26,7 +26,7 @@
       let commandIndex = $index
     ) {
       @if (command.recent && commandIndex === 0) {
-        <div class="list-group-item message message-md py-1 fw-bold">Recent</div>
+        <div class="list-group-item list-group-header">Recent</div>
       }
 
       <button
@@ -45,7 +45,7 @@
       </button>
 
       @if (command.recent && !commands[commandIndex + 1]?.recent) {
-        <div class="list-group-item message message-md py-1 fw-bold">Other Commands</div>
+        <div class="list-group-item list-group-header">Other Commands</div>
       }
     } @empty {
       <div class="list-group-item message message-md">No matching results</div>
--- a/packages/app/src/renderer/styles/boostrap-overrides.scss
+++ b/packages/app/src/renderer/styles/boostrap-overrides.scss
@@ -226,6 +226,12 @@ a:focus {
 .list-group-item {
   outline: none;
 }
+.list-group-header {
+  padding: 10px 10px 5px 10px;
+  color: $gray-600;
+  font-size: $font-size-sm;
+  font-weight: $font-weight-bold;
+}
 
 // inline code
 code {


packages/app/src/renderer/app/components/modals/command-palette-modal/command-palette-modal.component.html
```diff
@@ -35,6 +43,10 @@
           <span class="ms-auto" [innerHTML]="command.shortcut$ | async"></span>
         }
       </button>
+
+      @if (command.recent && !commands[commandIndex + 1]?.recent) {
+        <div class="list-group-item message message-md py-1 fw-bold">Other Commands</div>
+      }
```
255kb: Aren't the headers too big? I know there is no "list-group-header" in bootstrap but I think a design closes to the current dropdown headers would be better: 

Image

It could be doable with some classes like `list-group-item fw-medium ps-2 text-gray` (but text-gray would require to add `gray: $gray-600` to the `$theme-colors` list in bootstrap-variables.scss. 
You could also create a list-group-header class, mimicking the dropdown header. It may be easier.

255kb (changes requested):
Thanks @rohanakode490, it works great! 
A small remark regarding the headers. I think they should follow the dropdown's headers design if possible.

rohanakode490:
Like suggested. I went ahead with creating a `list-group-header` class for this.

255kb:
> Like suggested. I went ahead with creating a `list-group-header` class for this.

This looks great, thanks for the changes!

PR #2112, opened 2026-03-05
fix: ensure stringify handles nested SafeStrings

### LINKED ISSUE #2060: stringify produces unexpected JSON when object contains helper results (e.g. base64)
**Describe the bug**

When using `stringify` on an object that contains values produced by some helpers (e.g. `base64`), the output is not a JSON string value. Instead, it gets serialized as an object like `{ "string": "..." }`.

This makes the JSON output inconsistent depending on whether the value is passed through another string helper (e.g. `padStart`) or not.

**Steps to Reproduce**

1. Create a route (e.g. `GET /test`).
2. Response body (templating enabled), set the body to:

[template]
```hbs
{{{
  stringify (object
    foo = 'abc'
    bar = (base64 'abc')
    baz = (padStart (base64 'abc') 0 '')
  )
}}}
```

[output]
```json
{
  "baz": "YWJj",
  "bar": {
    "string": "YWJj"
  },
  "foo": "abc"
}
```

**Mockoon version:**

9.4.0

**OS / OS version:**

macOS

rohanakode490:

The stringify helper currently uses `JSON.stringify(data, null, 2)`. When helpers like `base64` or `padStart` are used inside an `(object ...)` helper, they return Handlebars SafeString objects.

While `JSON.stringify` works for primitive strings at the top level, it treats nested `SafeString` as regular JavaScript objects during serialization, resulting in an unexpected `{"string":"..."}` output for nested values.

I have updated the `stringify` helper to use a JSON replacer function. This ensures that any `SafeString` encountered during the recursive serialization process (at any nesting depth) is correctly converted back to a primitive string using the existing `fromSafeString` utility.

**Checklist**

Closes #2060

### DIFF
--- a/packages/commons-server/src/libs/templating-helpers/helpers/stringify.ts
+++ b/packages/commons-server/src/libs/templating-helpers/helpers/stringify.ts
@@ -12,7 +12,17 @@ const stringify = function (data: unknown, options: HelperOptions) {
       data = fromSafeString(data);
     }
 
-    return JSON.stringify(data, null, 2);
+    return JSON.stringify(
+      data,
+      (_, value) => {
+        if (value instanceof SafeString) {
+          return fromSafeString(value);
+        }
+
+        return value;
+      },
+      2
+    );
   } else {
     return data;
   }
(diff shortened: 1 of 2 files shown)


rohanakode490:
No problem :blush:

PR #2154, opened 2026-04-02
fix: add route metadata response headers #1837

### LINKED ISSUE #1837: Add response headers from the route with information about it
## Preface
Idea from this discussion https://github.com/mockoon/mockoon/discussions/1835#discussioncomment-14550380

## Problem
Currently, when creating a mocked route, it is not possible to put a header that would automatically take data about the route, for example, its id and description. This need arises when there are many routes to the same url with different data in mockoon and it is not convenient to search for which route gave the answer by logs or by the list of routes. It would be more convenient to have a response header that would immediately give an answer to which route it is from mockoon.

## Solution
As suggested in discussion https://github.com/mockoon/mockoon/discussions/1835#discussioncomment-14550380
It would be convenient to have such headers for every route.
```
x-mockoon-route-url: /path
x-mockoon-route-method: PATCH
x-mockoon-route-description: Hello
x-mockoon-route-uuid: xxxx-...
x-mockoon-route-response-no: 2
x-mockoon-route-response-uuid: xxxx-...
```

Brian-code-123:
## Summary
This PR adds automatic route metadata response headers for matched HTTP/CRUD routes in @mockoon/commons-server. 
Closes #1837

The following headers are now added to every served route response:
- x-mockoon-route-url
- x-mockoon-route-method
- x-mockoon-route-description
- x-mockoon-route-uuid
- x-mockoon-route-response-no
- x-mockoon-route-response-uuid

## Why
When multiple routes or multiple responses target similar endpoints, these headers make it much easier to identify exactly which route/response produced the payload during debugging and log analysis.

## Implementation details
- Added setRouteMetadataHeaders(...) in MockoonServer and called it in routeHandler(...) after custom response headers are set.
- x-mockoon-route-url uses the prepared/original route path (including endpoint prefix normalization).
- x-mockoon-route-method is exposed in uppercase.
- x-mockoon-route-response-no is the 1-based index of the selected response in the route responses list.

## Tests
- Added integration test: packages/commons-server/test/specs/server/route-metadata-headers.test.ts.
- Verified all six headers and values on a real server request.
- Ran full commons-server test suite successfully.

## Verification run
- npm run build:libs
- cd packages/commons-server && node --import tsx --test-concurrency 1 --test ./test/specs/server/route-metadata-headers.test.ts
- cd ../.. && npm run test:commons-server
- npm run lint --workspace=@mockoon/commons-server -- src/libs/server/server.ts test/specs/server/route-metadata-headers.test.ts

## AI disclosure
AI assistance was used to draft and implement this change, and the code/tests were reviewed and validated manually.


Brian-code-123:
Hi maintainers, could you please approve the workflow run for this PR? Thanks! 🙏  🙏

Brian-code-123:
## CI/CD Fix: Response Header Assertion Tests

### Problem
The Desktop packaged app E2E tests were failing on Linux and Windows due to the new route metadata headers added in this PR. The environment logs UI now shows additional `x-mockoon-route-*` headers alongside the response headers, which shifted the position of existing headers in the rendered list.

### Root Cause
The existing test assertions used hardcoded line indices to verify specific response headers (e.g., `assertLogItem(..., 'response', 4, 8)` checked item 8 in section 4). When the new route metadata headers were inserted, all subsequent headers shifted down in the list, causing these position-dependent assertions to look for the wrong header names.

### Solution
Implemented a more robust assertion approach:

1. **Added `assertLogSectionContains()` helper** in `environments-logs.ts`
   - Queries all items within a response/request section
   - Checks for header presence by content matching instead of position
   - Uses `.toContain()` to verify expected headers exist anywhere in the section

2. **Updated 3 proxy test assertions** in `proxy.spec.ts`
   - Replaced position-based assertions with the new content-based approach
   - Tests now verify that required headers are present, regardless of their position
   - Makes tests more resilient to future header additions

### Files Modified
- `packages/app/test/libs/environments-logs.ts` (+15 lines)
- `packages/app/test/specs/proxy.spec.ts` (+7 lines)

This fix allows the route metadata headers feature to work correctly while maintaining full E2E test coverage. The tests are now more robust and will continue to pass even if additional headers are added in the future.

255kb:
Thank you for contributing @Brian-code-123 
Could you please comment on the parent issue so I can assign it to you? We require contributors to be assigned to an existing issue prior to working on it.

Brian-code-123:
> Thank you for contributing @Brian-code-123 Could you please comment on the parent issue so I can assign it to you? We require contributors to be assigned to an existing issue prior to working on it.

Of course! Thanks for taking the time to read this PR. I've already left a comment on the parent issue.🙏🏻

Brian-code-123:
Hi maintainers, could you please approve the workflow run for this PR? I accidentally misclicked a button earlier. Thanks! 🙏🙏

255kb:
Now that I look at what I suggested in the parent issue/discussion I'm not so sure adding that many headers has value. 
I would like to avoid two things: clutter the responses with headers and add headers with content of variable size like the description. 

I would suggest the following changes: 

- remove the `x-mockoon-route-description` because of the potential bandwidth overhead.
- remove both `x-mockoon-route-url` and `x-mockoon-route-method` which seems redundant with the request payload. 
- make this configurable using a [server option](https://github.com/mockoon/mockoon/blob/3d0011391245a7e15a9459e9d9dfd4a725fd8698/packages/commons/src/models/server.model.ts#L88) and only append the header if it's activated.

255kb (approved):

PR #2160, opened 2026-04-08
fix: Retain literal boolean values in OpenAPI schemas

### LINKED ISSUE #2159: OpenAPI converter does not retain example/default false values
When importing an OpenAPI spec, if a `SchemaObject` is a `boolean` type and has an `example` or `default` value of `false`, it is not returned as the literal value in the data format and instead is replaced by a faker template. When the value is `true` it does not replace it, like the other data types.

Example schema YAML:

```yaml
openapi: '3.0.0'
info:
  title: 'example'
  version: '1.0.0'
paths:
  '/':
    get:
      responses:
        200:
          description: 'ok'
          content:
            'application/json':
              schema:
                type: 'object'
                properties:
                  false_default:
                    type: 'boolean'
                    default: false
                  true_default:
                    type: 'boolean'
                    default: true
                  false_example:
                    type: 'boolean'
                    example: false
                  true_example:
                    type: 'boolean'
                    example: true
```

Results in this JSON output:

```json
{
  "false_default": {{faker 'datatype.boolean'}},
  "true_default": true,
  "false_example": {{faker 'datatype.boolean'}},
  "true_example": true
}
```

It would be preferrable for the handling to be consistent with other data types, resulting in:

```json
{
  "false_default": false,
  "true_default": true,
  "false_example": false,
  "true_example": true
}
```

It is caused by a truthy check on these properties rather than `undefined` so I'm logging the issue before PR as it is a trivial fix. I hope this is okay!

**Mockoon version:**

\<= 9.6.1

**OS / OS version:**

Not Applicable

akamike:
When importing an OpenAPI spec, if a `SchemaObject` is a `boolean` type and has an `example` or `default` value of `false`, it is not returned as the literal value in the data format and instead is replaced by a faker template.

When the value is `true` it does not replace it. This is a small fix to replace the truthy check on these properties with an explicit `undefined` comparison so supplied values are consistently parsed.

**Checklist**

Closes #2159

### DIFF
--- a/packages/app/test/data/res/import-openapi/references/petstore-v2.json
+++ b/packages/app/test/data/res/import-openapi/references/petstore-v2.json
@@ -509,7 +509,7 @@
       "responses": [
         {
           "uuid": "334ea3fc-0c8d-4a07-b253-a51abbc612b9",
-          "body": "{\n  \"id\": {{faker 'number.int' max=99999}},\n  \"petId\": {{faker 'number.int' max=99999}},\n  \"quantity\": {{faker 'number.int' max=99999}},\n  \"shipDate\": \"{{faker 'date.recent' 365}}\",\n  \"status\": \"{{oneOf 'placed' 'approved' 'delivered'}}\",\n  \"complete\": {{faker 'datatype.boolean'}}\n}",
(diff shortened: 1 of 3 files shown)


255kb:
Thanks for the fix!

255kb (approved):

